PR #1, opened 2015-06-12
Added a global datadog service to evaluate the service

vnadgir-ft:
Have added a datadog service which is another monitoring solution amongst the mix.


Zapadlo:
Tried it already, closed service you have to pay for, provides basic metrics same as diamond or cadvisor.

PR #8, opened 2015-06-25
Expose aggregate-healthcheck to ELB

jgjay:
Will route all requests going through the ELB to aggregate-healthcheck by default. Overriding the Host header will allow you to get to anything else though.


jgjay:
How do we get the hostname of the ELB?

PR #13, opened 2015-06-29
Reg dep

Zapadlo:


martingartonft:
+1

[I don't really like all the things knowing about registry all over the place, but we can clean it up later.  I can't think of a better way right now that's simple.]

Zapadlo:
Same, I thought about this, I thought may be we can have a soft `PartOf=docker.service` and `Before=docker.service` but I'm not sure if thats better or even works. I think this is actually better, as each service can be explicit. I don't like forced caching.

PR #29, opened 2015-07-14
Switch deployer to run on a timer

Zapadlo:
https://github.com/Financial-Times/coco-fleet-deployer/pull/8

### DIFF
--- a/service-files/deployer.service
+++ b/service-files/deployer.service
@@ -5,6 +5,8 @@ Requires=docker.service
 
 [Service]
 TimeoutStartSec=600
+Type=oneshot
+RemainAfterExit=yes
 ExecStartPre=-/usr/bin/docker kill %p-%i
 ExecStartPre=-/usr/bin/docker rm %p-%i
 ExecStartPre=/usr/bin/docker pull coco/coco-fleet-deployer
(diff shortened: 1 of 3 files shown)


service-files/deployer.service
```diff
@@ -5,6 +5,8 @@ Requires=docker.service
 
 [Service]
 TimeoutStartSec=600
+Type=oneshot
+RemainAfterExit=yes
```
martingartonft: what does remain after exit do?

service-files/deployer.timer
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Deployer timer
+
+[Timer]
+OnCalendar=*-*-* *:*:00
```
martingartonft: how often is this?

service-files/deployer.service
Zapadlo: Service still looks "green" in systemctl and "active exited" instead of "innactive". No difference in functionality.

service-files/deployer.timer
Zapadlo: Every minute

martingartonft:
lgtm

PR #34, opened 2015-07-15
Changed to __health

Zapadlo:
Can get rid of the rewrite rule and is more inline with the standard


martingartonft:
lgtm, go, go, go!

tosan88:
+1

PR #38, opened 2015-07-17
route every request under the appropriate path

Nesze:


Nesze:
mainly for accessing the /history endpoint through elb

Zapadlo:
+1
makes sense to me

tosan88:
+1

PR #39, opened 2015-07-17
Sidekicks to ReloadPropagatedFrom main services

Zapadlo:
This fixes the bug where once main service is stopped and sidekicks stop
along with it, once main service it restarted, sidekick remains dead.
ReloadPropagatedFrom ensures that the reload of the main service is
propogated to the sidekick


martingartonft:
lgtm.

PR #47, opened 2015-08-03
System healthcheck

Nesze:


service-files/system-healthcheck.service
```diff
@@ -0,0 +1,22 @@
...
+Requires=docker.socket
+After=docker.socket
+
+[Service]
+Environment="DOCKER_APP_VERSION=latest"
+Environment="PROC_DIR=host_proc"
+ExecStartPre=-/usr/bin/docker kill %p-%H
+ExecStartPre=-/usr/bin/docker rm %p-%H
+ExecStartPre=/usr/bin/docker pull up-registry.ft.com/coco/system-healthcheck:$DOCKER_APP_VERSION
+ExecStart=/bin/bash -c "/usr/bin/docker run --rm --name=%p-%H -e PROC_DIR=$PROC_DIR -v /:/$PROC_DIR:ro up-registry.ft.com/coco/system-healthcheck:$DOCKER_APP_VERSION"
```
Zapadlo: You don't need to have %H in the container name, and considering how picky it is about characters, I think its actively better not to.

catalin-ilea:
lgtm

service-files/system-healthcheck.service
```diff
@@ -0,0 +1,22 @@
...
+Requires=docker.socket
+After=docker.socket
+
+[Service]
+Environment="DOCKER_APP_VERSION=latest"
+Environment="PROC_DIR=host_proc"
+ExecStartPre=-/usr/bin/docker kill %p
+ExecStartPre=-/usr/bin/docker rm %p
+ExecStartPre=/usr/bin/docker pull up-registry.ft.com/coco/system-healthcheck:$DOCKER_APP_VERSION
+ExecStart=/bin/bash -c "/usr/bin/docker run --rm --name=%p -e PROC_DIR=$PROC_DIR -v /:/$PROC_DIR:ro up-registry.ft.com/coco/system-healthcheck:$DOCKER_APP_VERSION"
```
Zapadlo: You are mounting ALL of host filesystem into the container, probably overkill and should only mount /proc:

``.. -v /proc:/$PROC_DIR:ro

Nesze: I did that on purpose, I thought for disk space check, I have to mount the whole fs.

Zapadlo: Right, makes sense.

Zapadlo:
+1

PR #60, opened 2015-08-15
Restart sidekicks when main service starts

Zapadlo:
I've done some testing on this locally, WantedBy on sidekick does not
work for this, the dependency MUST be specified on the main service to
work

@larisalungu I've tried this locally several times, but pleasse double check as I was sure I fixed this in the past and it turned out it wasn't working :P, ta!


larisalungu:
+1

Nesze:
+1

PR #64, opened 2015-08-20
Timeouts back 0 and killmode=none

Zapadlo:
The timeouts just cause pain and needless restarts, plus DO has theirs
at 0. killmode=none is again according to DO is needed to allow docker
--rm to work


Nesze:
+1

PR #68, opened 2015-08-26
Update mongo conf paths/ttls

Zapadlo:


catalin-ilea:
All services that read/write from mongo need updating as well.

Zapadlo:
Thanks, updated

catalin-ilea:
+1

PR #69, opened 2015-08-26
Externalise splunk forwarder images to dockerhub

Zapadlo:
This doesn't work being in up-registry and registry itself uses the same
image. Reason its angelx and not coco, is that there is currently an
issue with dockerhub that is preventing us from deleting the coco
automated build. And the build won't work. So temp work around, putting
it on my account


catalin-ilea:
+1

PR #70, opened 2015-08-27
Pre prod

Zapadlo:


catalin-ilea:
+1

martingartonft:
is this really config?  /services/foo sucks and /ft/something/foo is the way to do.

but this seems more like application state, not config.  (like /var not /etc)

Zapadlo:
Happy for this to live elsewhere, just not the ambiguous `/services`..

PR #79, opened 2015-09-09
Removed host name from fleet-unit-hc routing rule

Nesze:
This also fixes the problem when fleet-unit-hc moves to other machine, and agg-hc continues to look for fleet-unit-hc on the old machine as well.


Zapadlo:
+1

PR #82, opened 2015-09-09
first try to propagate restart to Kafka dependent servicess

Nesze:
1. listing these services in Kafka's service file PropagatesReloadTo section would mean to list all the instances of the services, which is not ideal  
2. I've also taken a look at type=notify services, but that means lot of changes (I think it's safe to say that it's not meant for this).


kovacshuni:
+1

Zapadlo:
Could you test this by restarting Kafka and see if it works?

Nesze:
that's what I intend, note that is a PR to pre-prod, I might break it a bit

Zapadlo:
OK, go for it then

PR #85, opened 2015-09-14
trying out PartOf option for propagating restart

Nesze:


Nesze:
I'm going to merge this without approval : /

Zapadlo:
I think this is similar scenario to our sidekicks, niether of these options will work. Something needs to live on the kafka service. I don't know how to solve the "multiple instances" problem, can we use wildcards?

Nesze:
Right, gonna revert it, doesn't seem to work.

PR #89, opened 2015-09-21
Give docker container a name, so make sure it gets removed

Zapadlo:
Also delete the image for docker-gc on fail, image-cleaner fixes this
problem for the other services, but if its affected by it itself, it
fails


Nesze:
+1

PR #91, opened 2015-09-22
Fork docker-gc while spotify image is broken

Zapadlo:


martingartonft:
lgtm. maybe semi-permanent fork which we re-sync if needed?

Zapadlo:
May be, but it can easily become a thing we forget about...

PR #93, opened 2015-09-24
Upgrade to a version that is based on new dwbase (pre-prod)

martingartonft:
This uses a new dwbase in dockerhub which has no ft deps in it. Also, we
use official alpine openjdk8.


Zapadlo:
+1

catalin-ilea:
+1

Nesze:
+1

PR #103, opened 2015-09-28
Bump notifications-reader version; listen on ports 8080/8081.

khatton-ft:
Dockerfile in notifications-reader-service now forces the app to listen on 8080/8081 and reads MongoDB locations from $MONGO_ADDRESSES.


Zapadlo:
+1

PR #107, opened 2015-10-02
Remove healthcheck endpoint key

Nesze:
Reason: the service name contains the hostname, so whenever a box is replaced in the cluster, agg-hc still keeps listing the deleted box' system healthcheck.
In case of system-healthcheck gets inactive, fleet-unit-healthcheck still catches the problem.


martingartonft:
tentative +1.  would be good to hear what others think.

Zapadlo:
+0.5 I think its good, I'm a little uneasy we won't know if a box just drops out and our monitoring is all green. Def bad.

martingartonft:
Yes, but if you have the mindset that machines appearing and disappearing is normal, and you rely on other health checks for functionality, then all this one is for really is to ensure the machines that _are_ there are happy.

Zapadlo:
Yes, what I want is *amber* before *red*. I want to avoid *green* to *red*. Box drops, we are still green, second box drops, we are still green, third box drops and everything breaks. I want to be notified of upcoming failure due to present happennings. Whatever that mechanism might be.

martingartonft:
We can monitor how many of the machines in the cluster (discovery url) are actually present?
Any missing == amber. half or fewer == red?

Zapadlo:
I like that, we can use `etcdctl cluster-health` to monitor for missing machines. It might bleep when we restart, but will recover when it comes up. I don't think we should **red** anything thats not a business function check. If its somehow working on one machine - fine.

Nesze:
This would be genuinely useful.  

I park this PR until we add that extra `etcd-cluster-health` service.

Zapadlo:
May be it can be part of https://github.com/Financial-Times/coco-fleet-unit-healthcheck? In my mind we need two services, one per box, that check CPU/memory/disk and one per cluster that checks services, etcd health, etc.

martingartonft:
agreed. fleet-unit-healthcheck could become fleet-cluster-healthcheck

Zapadlo:
Finally, I was expecting three comments instead trashin my idea in a google doc.... traitor

PR #110, opened 2015-10-06
Running mongo configurator only once.

catalin-ilea:


tamas-molnar:
+1

catalin-ilea:
Fair point...nothing atm. Will look into it.

PR #113, opened 2015-10-13
Lowering backup process priority

tamas-molnar:
* I chose 10 as a middle-ground, I'll monitor the next backup and adjust if necessary
* I/O level adjustments shouldn't be necessary as the archived folder is streamed into gof3r and on to S3, it doesn't stick around much, if at all, on disk


Zapadlo:
+1

catalin-ilea:
+1

PR #114, opened 2015-10-14
Fix backup timer

tamas-molnar:
- the timer has to live on the same machine as the service it's starting


Zapadlo:
This kind of PRs is the reason we should not have branch > pre-prod > prod progression.

PR #119, opened 2015-10-16
Adding mongo-backup service, timer and service definition

tamas-molnar:
* same service, timer and definition as on the `pre-prod` branch.
* I left the schedule the same as in `pre-prod` so we're here to see it happen. AFAIK we will only run this for a few days anyway. 
* added necessary data in `etcd` in prod as well

### DIFF
--- a/service-files/mongo-backup.timer
+++ b/service-files/mongo-backup.timer
@@ -1,0 +1,12 @@
+[Unit]
+Description=Runs the mongo backup job
+
+[Timer]
+OnCalendar=11:00
+
+[Install]
+WantedBy=timers.target
+
+[X-Fleet]
+MachineOf=mongo-backup.service
+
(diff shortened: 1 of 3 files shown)


catalin-ilea:
lgtm

service-files/mongo-backup.timer
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Runs the mongo backup job
+
+[Timer]
+OnCalendar=11:00
```
Zapadlo: Do you also want: http://www.freedesktop.org/software/systemd/man/systemd.timer.html#Persistent=?

Zapadlo: Do you also want: http://www.freedesktop.org/software/systemd/man/systemd.timer.html#Persistent=?

tamas-molnar: No, because that might mean that it gets run every time it's moved from one machine to the other in case of dissapearing instances for example.

Zapadlo: Yep, fine.

PR #127, opened 2015-10-21
Using docker rm -f instead of docker stop to remove containers

catalin-ilea:


martingartonft:
lgtm. "merge, merge, merge!"

PR #134, opened 2015-10-28
Unschedule skydns, currently not used by anything

Zapadlo:
Leaving the unit file in, in case we want to use it for anything in the
future


martingartonft:
lgtm. bet you are slightly sad though..  don't worry, i bet it will return.

PR #139, opened 2015-11-03
added vulcan-config-builder

martingartonft:
Added vulcan-config-builder service. Currently nothing uses it.

### DIFF
--- a/vulcan-config-builder.service
+++ b/vulcan-config-builder.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Vulcand configuration builder.
+Requires=docker.socket
+After=docker.socket
+
+[Service]
+ExecStartPre=-/usr/bin/docker kill %p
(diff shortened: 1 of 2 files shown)


vulcan-config-builder.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Vulcand configuration builder.
+Requires=docker.socket
```
Zapadlo: I would add etcd2.service and vulcan.service here.

Zapadlo:
+1

vulcan-config-builder.service
martingartonft: vcb doesn't need vulcan running to set the values in etcd so vulcan can start after vcb just fine.

Zapadlo: OK, nvm

PR #141, opened 2015-11-04
added restorage instance running on mongo

martingartonft:
This adds an instance of up-restorage to production, configured to point at mongodb.  Among other things, this gives a nice way to stream individual collections of data out of mongodb via curl, which can then (for example) be piped into other clusters, or into other restorage backends (e.g, elasticsearch)

Note that this app is single instance and not explicitly monitored etc at the moment as it's currently more of an operational tool than anything else.


catalin-ilea:
+1

PR #144, opened 2015-11-05
Pre prod

catalin-ilea:


Zapadlo:
+1

Nesze:
+1

Are these recognized by the proxy?

catalin-ilea:
@Nesze Asking this after a +1 ? :)
/usr/local/bin/rest-proxy-docker.sh

for var in $(env | grep -v '^RP_CFG_' | grep '^RP_' | sort); do
  key=$(echo $var | sed -r 's/RP_(.*)=.*/\1/g' | tr A-Z a-z | tr _ .)
  value=$(echo $var | sed -r 's/.*=(.*)/\1/g')
  echo "${key}=${value}" >> ${rp_cfg_file}
done

catalin-ilea:
So answer is, yes they are if you follow the right pattern

PR #145, opened 2015-11-05
Convert native rw to vcb

martingartonft:
tested in my own cluster. ready for pre-prod at least I think,


Zapadlo:
Very funky +1

BerniVarga:
+1

PR #171, opened 2015-11-26
In diamond to send metrics, using only the part of ip-172-24-75-233 f…

kovacshuni:
…rom hostname ip-172-24-55-177.eu-west-1.compute.internal or ip-172-24-75-233.ec2.internal to make it have the same format for every environment.


Zapadlo:
+1

catalin-ilea:
+1

PR #174, opened 2015-11-30
Convert to use vulcan config builder

Zapadlo:


martingartonft:
have you tested this anywhere yet?  It's not a numbered templated service and I don't think we've tried these yet?

Zapadlo:
Nope.... :)

martingartonft:
bet you 5p it doesn't work.

PR #176, opened 2015-12-01
Explicitly turn off docker logging for filter and http-forwarder

Zapadlo:
As it currently stands stdout is logged to a json file in
`/var/lib/docker/containers` and can consume all FS space. Logs are
already going into journald by virtue of process being wrapped in
`systemd` these however are managed and will not consume ALL space.


martingartonft:
lgtm

PR #179, opened 2015-12-02
Standardizing mongo backup config

tamas-molnar:
* read the bucket, data dir, and s3 domain from the service file instead of etcd
* the data dir and s3 domain are the same in all environments, so no reason to have them in etcd
* to simplify the configuration, we can keep all backups in the same bucket, as the backup file names are suffixed with the environment
* otherwise, to automate things, we would either have to dynamically build the bucket name in the provisioner
 * we currently have 1 bucket for pre-prod, 1 bucket for prod, and 2 prod envs
 * the other option is to manually add the values after the provisioner is done, which is not desirable


kovacshuni:
+1

PR #182, opened 2015-12-04
Standardize Mongo Backup Config

tamas-molnar:
* tried it in pre-prod, backup works with the values hard-coded in the service file


Nesze:
lgtm

PR #201, opened 2015-12-17
Merge changes from master to pre-prod cluster

BerniVarga:


khatton-ft:
Looks good but I would suggest picking up changes from #202 (when approved) to update the notifications ingester.

PR #210, opened 2015-12-17
synthetic-request-monitor coco to supply authorization header when ma…

kovacshuni:
…king requests to cms-notifier


catalin-ilea:
lgtm

PR #227, opened 2016-01-08
Sorted deduped dp services file

Zapadlo:


martingartonft:
lgtm

khatton-ft:
+1 I notice WPAT wasn't the only one with duplicate entries ... I wonder if that is why this cluster has been problematic.

PR #238, opened 2016-01-11
Updated enriched content read API services files

tamas-molnar:
* switched Enriched Content Read API to VCB config
* updated service files to latest conventions
* added app to services.yaml
* this isn't expected to run/be merged, just a trial version


tosan88:
:+1:

PR #239, opened 2016-01-11
promote pre-prod to prod.

kovacshuni:


Zapadlo:
+1

tosan88:
That's good, because you get rid of the varnish usage in `services.yaml` :)
:+1:

PR #245, opened 2016-01-13
Make sure we delete the node before creating a record

Zapadlo:
Konstructor will append a record to the node if one already exists. What
we want is to guarantee replacement. This will make the service fail if
we were unable to delete AND create an entry, making the operation
atomic


Nesze:
+1

tosan88:
:+1:

tamas-molnar:
+1

martingartonft:
lgtm

PR #246, opened 2016-01-13
neo services

martingartonft:
neo services (hand copied from elasticSearch branch)

NOT USED (yet)


Zapadlo:
Pinned port?

martingartonft:
probably not for long, but yes for the moment.  I can't remember the issues, but I started not pinned and then pinned it.

Zapadlo:
+1

PR #256, opened 2016-01-19
Containerize api comp policy

BerniVarga:


notifications-rw-sidekick@.service
```diff
@@ -8,6 +8,7 @@ ExecStart=/bin/sh -c "\
   etcdctl set ft/services/notifications-rw/healthcheck true; \
   etcdctl mkdir /ft/services/notifications-rw/servers; \
   etcdctl set /ft/healthcheck/notifications-rw-%i /__health; \
+  etcdctl set /vulcand/frontends/public-services-notifications/frontend '{\"Type\": \"http\", \"BackendId\": \"vcb-notifications-rw\", \"Route\": \"PathRegexp(`/notifications.*`) && Host(`public-services`)\"}'; \
```
khatton-ft: Shouldn't this be a change to vulcan-config-builder instead?

document-store-api-sidekick@.service
```diff
@@ -8,6 +8,8 @@ ExecStart=/bin/sh -c "\
   etcdctl set   /ft/services/document-store-api/healthcheck     true; \
   etcdctl mkdir /ft/services/document-store-api/servers; \
   etcdctl set /ft/healthcheck/document-store-api-%i /__health; \
+  etcdctl set /vulcand/frontends/public-services-content/frontend '{\"Type\": \"http\", \"BackendId\": \"vcb-document-store-api\", \"Route\": \"PathRegexp(`/content/.*`) && Host(`public-services`)\"}'; \
+  etcdctl set /vulcand/frontends/public-services-lists/frontend '{\"Type\": \"http\", \"BackendId\": \"vcb-document-store-api\", \"Route\": \"PathRegexp(`/lists/.*`) && Host(`public-services`)\"}'; \
```
khatton-ft: Shouldn't this be a change to vulcan-config-builder? It looks like we're proliferating config again here.

api-policy-component@.service
```diff
@@ -5,17 +5,20 @@ Requires=docker.service
...
 KillMode=none
 ExecStartPre=-/bin/bash -c '/usr/bin/docker kill %p-%i > /dev/null 2>&1'
 ExecStartPre=-/bin/bash -c '/usr/bin/docker rm %p-%i > /dev/null 2>&1'
-ExecStartPre=/usr/bin/docker pull up-registry.ft.com/coco/api-policy-component
+ExecStartPre=/bin/bash -c 'docker pull up-registry.ft.com/coco/api-policy-component:$DOCKER_APP_VERSION'
 
-ExecStart=/bin/sh -c "\
-/usr/bin/docker run --rm --name %p-%i -p 8080 -p 8081 --env \"READ_ENDPOINT=$HOSTNAME:8080\" up-registry.ft.com/coco/api-policy-component;"
+ExecStart=/bin/sh -c '\
+/usr/bin/docker run --rm --name %p-%i -p 8080 -p 8081 --env "READ_ENDPOINT=$HOSTNAME:8080:8182" up-registry.ft.com/coco/api-policy-component:$DOCKER_APP_VERSION'
```
khatton-ft: 8182? I think it's worth a comment in the file to explain that port number.

Zapadlo:
Your branch has an outdated services.yaml with a number of duplicates that since been removed, you might want to catch this branch up

notifications-rw-sidekick@.service
BerniVarga: This would mean that the vulcan-config-builder should generate a new frontend to the actual service's backend, containing a routing path and a host header - both given as parameters (these values have to come from outside, can't they just be automatically generated). As far as I know, there is no functionality in the vulcan-config-builder which adds a different host header that the default one to a frontend. Are suggesting adding something like that?

document-store-api-sidekick@.service
Zapadlo: vcb currently covers default functionality like, 1 FE, 1 BE per service, it can't do much more then that. So currently "custom" rules are done directly in vulcan config

api-policy-component@.service
BerniVarga: You are right. I will add a comment. Through this port the vulcan api can be reached.

PR #257, opened 2016-01-19
Varnish routing

tosan88:
Varnish forwards requests to api-policy-component (instead of document-store-api).
Now varnish routing rule for /content and /lists endpoint are added by document-store-api sidekick.


tamas-molnar:
+1

PR #260, opened 2016-01-20
Pre prod

Nesze:
Most recent Wordpress-Article-Transformer with some fixes
+
some neo4j stuff

Edit: Question to you all, should we set the sequentialDeployment flag to true? For _read_ services it's obvious, although, here I'm not 100% sure. It would make sense to set for _all_ the services then.


BerniVarga:
+1

tamas-molnar:
lgtm

PR #261, opened 2016-01-20
Merging into XP: pre-prod and containerize-api-comp-policy

BerniVarga:
A final version of the api-policy-comp containerization.


tosan88:
:+1:

PR #262, opened 2016-01-20
Containerize api-policy-component

BerniVarga:
Tested in xp.

### DIFF
--- a/notifications-rw-sidekick@.service
+++ b/notifications-rw-sidekick@.service
@@ -8,6 +8,7 @@ ExecStart=/bin/sh -c "\
   etcdctl set ft/services/notifications-rw/healthcheck true; \
   etcdctl mkdir /ft/services/notifications-rw/servers; \
   etcdctl set /ft/healthcheck/notifications-rw-%i /__health; \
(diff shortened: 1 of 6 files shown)


tamas-molnar:
+1

notifications-rw-sidekick@.service
```diff
@@ -8,6 +8,7 @@ ExecStart=/bin/sh -c "\
   etcdctl set ft/services/notifications-rw/healthcheck true; \
   etcdctl mkdir /ft/services/notifications-rw/servers; \
   etcdctl set /ft/healthcheck/notifications-rw-%i /__health; \
+  etcdctl set /vulcand/frontends/public-services-notifications/frontend '{\"Type\": \"http\", \"BackendId\": \"vcb-notifications-rw\", \"Route\": \"PathRegexp(`/notifications.*`) && Host(`public-services`)\"}'; \
```
Nesze: I wonder why is this working, "PathRegexp(`/notifications.*`)".
I thought it should match the whole path.

Nesze:
+1

notifications-rw-sidekick@.service
BerniVarga: I tried to find the answer in vulcan's code, but couldn't find it. Anyway, I changed the regex path as you suggested - to keep consistency.

martingartonft:
+1

PR #265, opened 2016-01-20
Fixed content ingester whitelist: now image models will be ingested (PROD)

izzyblues:
Fixed content ingester whitelist: now image models will be ingested


Zapadlo:
Do you mean to deploy neo4j/document-store-api/nativerw, etc as well?

The way merge from the dynpub branch can work if you first merge pre-prod into dynpub, that merge the PR into pre-prod, then rebase the dynpub off master and merge the PR into master. But that sounds like a fair amount of effort.

May I suggest to keep your feature changes on a feature branch ei `fixed-content-ingester-whitelist` and merge that into dynpub as you continue working on it as a way of testing it in an environment. Once its ready for deployment, create a PR from `fixed-content-ingester-whitelist` to `pre-prod` and a seperate one from `fixed-content-ingester-whitelist` to `prod`. That way only your changes will be present in the PRs and merges are significantly less likely.

Perhaps we need a different approach to that to fit your requirements, but 95 commit PRs are hard to review and can be dangerous.

izzyblues:
Hi George,
I agree with you.
I will close the two pull requests and work on a feature branch.

PR #269, opened 2016-01-21
Setup varnish for caching requests going to content and lists endpoints

tosan88:
Finally we managed to setup varnish as default entry point for requests in xp

### DIFF
--- a/varnish-sidekick@.service
+++ b/varnish-sidekick@.service
@@ -5,15 +5,17 @@ After=varnish@%i.service
 
 [Service]
 ExecStart=/bin/sh -c "\
-  export VARNISH_IP=$HOSTNAME; \
+  etcdctl set   /ft/services/varnish/healthcheck true; \
+  etcdctl mkdir /ft/services/varnish/servers; \
+  etcdctl set /ft/healthcheck/varnish-%i /__health; \
   while true; do \
-    export VARNISH_PORT=$(/usr/bin/docker port varnish-%i 80 | cut -d':' -f2); \
-    etcdctl set /services/varnish/%i/host $VARNISH_IP --ttl 60; \
-    etcdctl set /services/varnish/%i/port $VARNISH_PORT --ttl 60; \
+    export PORT=$(/usr/bin/docker port varnish-%i 80 | cut -d':' -f2); \
+    etcdctl set /ft/services/varnish/servers/%i http://$HOSTNAME:$PORT --ttl 600; \
     sleep 45; \
   done"
-ExecStop=/usr/bin/etcdctl rm /services/varnish@%i
+ExecStop=/bin/sh -c "\
+  etcdctl rm /ft/services/varnish/servers/%i; \
+  etcdctl rm /vulcand/frontends/varnish-public/frontend"
 
 [X-Fleet]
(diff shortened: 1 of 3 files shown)


varnish-sidekick@.service
```diff
@@ -5,15 +5,17 @@ After=varnish@%i.service
...
-    etcdctl set /services/varnish/%i/host $VARNISH_IP --ttl 60; \
-    etcdctl set /services/varnish/%i/port $VARNISH_PORT --ttl 60; \
+    export PORT=$(/usr/bin/docker port varnish-%i 80 | cut -d':' -f2); \
+    etcdctl set /ft/services/varnish/servers/%i http://$HOSTNAME:$PORT --ttl 600; \
     sleep 45; \
   done"
-ExecStop=/usr/bin/etcdctl rm /services/varnish@%i
+ExecStop=/bin/sh -c "\
+  etcdctl rm /ft/services/varnish/servers/%i; \
+  etcdctl rm /vulcand/frontends/varnish-public/frontend"
```
BerniVarga: Does this key still exist?

tosan88: No, you're right. I'll fix this

varnish-sidekick@.service
```diff
@@ -5,15 +5,16 @@ After=varnish@%i.service
...
-  export VARNISH_IP=$HOSTNAME; \
+  etcdctl set   /ft/services/varnish/healthcheck true; \
+  etcdctl mkdir /ft/services/varnish/servers; \
+  etcdctl set /ft/healthcheck/varnish-%i /__health; \
   while true; do \
-    export VARNISH_PORT=$(/usr/bin/docker port varnish-%i 80 | cut -d':' -f2); \
-    etcdctl set /services/varnish/%i/host $VARNISH_IP --ttl 60; \
-    etcdctl set /services/varnish/%i/port $VARNISH_PORT --ttl 60; \
+    export PORT=$(/usr/bin/docker port varnish-%i 80 | cut -d':' -f2); \
+    etcdctl set /ft/services/varnish/servers/%i http://$HOSTNAME:$PORT --ttl 600; \
```
Zapadlo: If you ttl is 600s you probably don't need to refresh it every 45s

Zapadlo:
I would have still preferred a less chatty approach with linking varnishes to APCs, but..

+1

BerniVarga:
+1 (Would it make sense to put a task of optimization on our board for varnish-apc coupling?)

varnish-sidekick@.service
tosan88: Ah, so what would fit better: to change TTL to be less (120s maybe) or to increase the sleep? or both? :)

tosan88:
@Zapadlo @BerniVarga I thought the same, we could implement linking as a further improvement

Zapadlo:
@tosan88 sgtm

varnish-sidekick@.service
Zapadlo: How about ttl of 5 and sleep of 4?

In line with this: https://github.com/Financial-Times/up-service-files/pull/271

tosan88: Updated.

PR #271, opened 2016-01-22
Implement short heartbeats for read services

Zapadlo:
This is part of the effort to create a read endpoint healthcheck which
would monitor availability of at least one service being available

### DIFF
--- a/api-policy-component-sidekick@.service
+++ b/api-policy-component-sidekick@.service
@@ -8,10 +8,10 @@ ExecStart=/bin/sh -c "\
   etcdctl set /ft/services/api-policy-component/healthcheck true; \
   etcdctl mkdir /ft/services/api-policy-component/servers; \
   etcdctl set /ft/healthcheck/api-policy-component-%i /__health; \
+  PORT=$(/usr/bin/docker port api-policy-component-%i 8080 | cut -d':' -f2); \
   while true; do \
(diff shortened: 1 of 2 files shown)


BerniVarga:
+1

martingartonft:
Do we want something more lightweight than __health?
Also, please hold off until we get to the bottom of the current mongodb connection issues.

Zapadlo:
@martingartonft We do, but we don't have.

Will hold off :)

kovacshuni:
+1

tosan88:
:+1:

martingartonft:
Can someone remind me what problem this one solves?  I'm nervous about calling healthchecks every few seconds, and also setting etcd value (which required cross-cluster consensus) just as often.

Zapadlo:
Vulcan routing to broken services.

Once the aggregate-healthcheck is able to report on the health of the read services, it will respond as "healthy" as long as an instance of each type of service is available. This means Dyn will route requests that can however fail if an application is running but is unhealthy.

martingartonft:
if we had __gtg, I'd prefer that.  We don't yet in those apps, so this LGTM.

Zapadlo:
Agreed, happy to switch to gtg as soon as thats available

Zapadlo:
Deployed to pre-prod

PR #275, opened 2016-01-29
Using varnish for requests having Host: read-$ENV-up.ft.com

tosan88:


tosan88:
I still need to add some more configuration, hold on

catalin-ilea:
lgtm

PR #290, opened 2016-02-02
Varnish

tosan88:
Varnish &
+ PAM version increment (commit from pre-prod not merged to XP)

### DIFF
--- a/varnish-sidekick@.service
+++ b/varnish-sidekick@.service
@@ -15,8 +15,7 @@ ExecStart=/bin/sh -c "\
     sleep 4; \
   done"
 ExecStop=/bin/sh -c "\
-  etcdctl rm /ft/services/varnish/servers/%i; \
-  etcdctl rm /vulcand/frontends/varnish-public-endpoint/frontend"
+  etcdctl rm /ft/services/varnish/servers/%i"
 
 [X-Fleet]
 MachineOf=varnish@%i.service
--- a/varnish-sidekick@.service
+++ b/varnish-sidekick@.service
@@ -5,15 +5,18 @@ After=varnish@%i.service
 
 [Service]
 ExecStart=/bin/sh -c "\
-  export VARNISH_IP=$HOSTNAME; \
+  etcdctl set   /ft/services/varnish/healthcheck true; \
+  etcdctl mkdir /ft/services/varnish/servers; \
+  etcdctl set /ft/healthcheck/varnish-%i /__health; \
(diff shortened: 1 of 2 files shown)


catalin-ilea:
lgtm

varnish-sidekick@.service
```diff
@@ -5,15 +5,17 @@ After=varnish@%i.service
 
 [Service]
 ExecStart=/bin/sh -c "\
-  export VARNISH_IP=$HOSTNAME; \
+  etcdctl set   /ft/services/varnish/healthcheck true; \
+  etcdctl mkdir /ft/services/varnish/servers; \
+  etcdctl set /ft/healthcheck/varnish-%i /__health; \
+  etcdctl set /vulcand/frontends/varnish-public-endpoint/frontend '{\"Type\":\"http\", \"BackendId\":\"vcb-varnish\", \"Route\":\"PathRegexp(`/.*`) && MethodRegexp(`GET|HEAD`) && HostRegexp(`read-.*-up.ft.com`)\"}'; \
```
Zapadlo: Need to escape dots in HostRegexp `read-.*-up\.ft\.com`

varnish-sidekick@.service
```diff
@@ -5,15 +5,17 @@ After=varnish@%i.service
...
   while true; do \
-    export VARNISH_PORT=$(/usr/bin/docker port varnish-%i 80 | cut -d':' -f2); \
-    etcdctl set /services/varnish/%i/host $VARNISH_IP --ttl 60; \
-    etcdctl set /services/varnish/%i/port $VARNISH_PORT --ttl 60; \
-    sleep 45; \
+    export PORT=$(/usr/bin/docker port varnish-%i 80 | cut -d':' -f2); \
```
Zapadlo: I think we should move this command outside the while loop. There is no way service can change the port on the machine within the lifecycle of the sidekick.

tosan88: I can move this one, but other services do the same thing.

varnish-sidekick@.service
tosan88: Right, I'm doing it :)

varnish-sidekick@.service
Zapadlo: Yep, we'll move then in due time, but this service is also one of few that sleep for only 4 seconds.

Zapadlo:
+1

martingartonft:
+1

PR #291, opened 2016-02-03
Deploy sipm with PID 1 fix

Zapadlo:
Deploy version of SIPM that runs the application process as PID 1 in
attempt to fix the issue with orphaned containers


tosan88:
:+1: 
But please create a normal version of the application if the fix works before going to master.

Zapadlo:
I will, currently the change is on a branch.

tosan88:
@Zapadlo Thanks! :)

PR #299, opened 2016-02-04
Register read dns

tosan88:
* Using new version of elb-dns-registrator to support multiple domain creation - $ENV-up (current) and $ENV-up-read (for read endpoints)

* Fixing varnish's Host regex
* Versions of doc store api and api policy component. Rename opening to openingXML and apply INCLUDE_RICH_CONTENT policy to openingXML
* Bump PAM version. Skip checking live content from notifications endpoint
* Deploy version of SIPM that runs the application process as PID 1 in
attempt to fix the issue with orphaned containers


catalin-ilea:
lgtm

PR #301, opened 2016-02-05
Pre prod

tosan88:
* Using new version of elb-dns-registrator to support multiple domain creation - $ENV-up (current) and $ENV-up-read (for read endpoints)
* Introducing varnish
* Bump PAM version. Skip checking live content from notifications endpoint
* <s>Deploy version of SIPM that runs the application process as PID 1 in attempt to fix the issue with orphaned containers</s> - reverted


Zapadlo:
+1

PR #302, opened 2016-02-05
Fix konstructor calls

Zapadlo:
Do a DELETE, disregard the result. Check there is no record (404)
followed by a create.


catalin-ilea:
lgtm

PR #305, opened 2016-02-08
Fix image related issues

tosan88:
* Addig whiteListRegex for binary-ingester to avoid processing non-images (binary-ingester v82)
* Fixing externalBinaryUrl for image models (methode-image-model-transformer v42)

* Revert "Deploy sipm with PID 1 fix"
* Fix konstructor calls a DELETE, disregard the result. Check there is no record (404)
followed by a create


tamas-molnar:
lgtm, but why are there so many changes that were in pre-prod and not in xp?

tosan88:
commits are in xp, only merge commits are not. Github doesn't show it correctly, but if you take a look on the xp branch, you can see the most changes are already there

BerniVarga:
+1

PR #306, opened 2016-02-09
Sequential deployment enabled for read services. Updated elb-dns-regi…

tosan88:
…strator version to fix small bug


Zapadlo:
I don't think its applicable to set SD (sequential deployment) for sidekicks

catalin-ilea:
Agree with @Zapadlo

Zapadlo:
+1

BerniVarga:
+1

PR #311, opened 2016-02-10
New SIPM service

Zapadlo:
New version with `exec` that fixes orphaned containers
New pull format to stop pull requests when image is present
Moved the images to dockerhub


BerniVarga:
+1

BerniVarga:
However... don't we want to stick to the pattern we agreed on for merging? (first merge the branch in pre-prod, then pre-prod in master... otherwise we will end-up with a lot of different merge hashes...)

Zapadlo:
Only if there are conflicts, but I'm happy to close this and just leave the master PR, or other way around

PR #319, opened 2016-02-11
Versioning vulcan

tosan88:
* Updated the version of the Methode-Article-Transformer to accept POST with payload


Zapadlo:
+1

PR #320, opened 2016-02-11
Read services sequential deployment

tosan88:


tamas-molnar:
+1

Zapadlo:
deja vu, I swear I approved a PR exactly like this... it had sequential deployment on sidekicks at first

tosan88:
@Zapadlo You're right, that was for xp branch :) I did not see any issues with that, so now I'm aiming pre-prod

PR #343, opened 2016-02-17
Additionally improved content-public-read-preview service file

izzyblues:
The content-public-read-preview service file contains the script to set the API host configuration according to the environment in which the service is running.


Zapadlo:
+1

khatton-ft:
:+1:

PR #355, opened 2016-02-19
Fixed missing PORT and LB member check

Zapadlo:
PORT getting is done inside the loop, but `docker port` only gets called
if $PORT is empty, once populated value is reused

Also checks healthchecks of the main service, if all are true, sets the
value in etcd


kovacshuni:
+1

tosan88:
:+1:

tamas-molnar:
Is `jq` included in CoreOs by default?

Zapadlo:
@tamas-molnar Yep, very happy about that

tamas-molnar:
+1

Zapadlo:
Deployed to xp

catalin-ilea:
Can we add the same thing on varnish?

Zapadlo:
The port - yes, it doesn't have healthchecks, does it?

catalin-ilea:
Check __health as well. It will hit api-component-policy/__health but it will tell us if varnish is running. Bad idea?

Zapadlo:
Yes... you are going to hammer ACP/__health every 2.5 seconds. Is varnish likely to run broken?

catalin-ilea:
Probaly not. Ok, fair point, leave it as is.

Zapadlo:
Deployed to `pre-prod`

PR #364, opened 2016-02-23
Deleting unused service files

tamas-molnar:
* The goal of this PR is to delete all unused service files
* I've been heavy-handed in deleting them, so if you think they should be kept, please drop a comment


Zapadlo:
The best kind of PR +1

PR #365, opened 2016-02-24
Sections

tosan88:
* Adding sections-transformer and sections-rw-neo4j services files
* Configured those services in services.yaml

Will be testing in XP cluster


tamas-molnar:
lgtm

tosan88:
All good in xp. Going to pre-prod

PR #385, opened 2016-03-01
Updating sections- and subjects-rw: using Classification category for…

tosan88:
… Neo interactions


tosan88:
Tested in xp and it works as expected

Zapadlo:
+1

tosan88:
Tested in pre-prod, works as expected

PR #396, opened 2016-03-03
Update public-annotations-api to v0.0.4 - returning v1 annotations.

BerniVarga:
This version will give us back correctly all the v1 annotations. (tested in xp)


catalin-ilea:
lgtm

PR #397, opened 2016-03-03
Adding genres-rw-neo4j

tamas-molnar:
* tried in XP, read and write working
* also check out https://github.com/Financial-Times/genres-rw-neo4j, adapted from the subjects counterpart


tosan88:
lgtm

Zapadlo:
+1

PR #412, opened 2016-03-09
Transformed payload on kafka

martinroddam:


khatton-ft:
I see we are going to pick up some other changes with this one so it may need a careful watch.
:+1:

Zapadlo:
Same, is this still relevant?

khatton-ft:
A later version of the content-importer has already been deployed.

PR #419, opened 2016-03-16
Limit ingester services to 512M memory

Zapadlo:


Nesze:
+1

kovacshuni:
+1

khatton-ft:
The ingester Dockerfile specifies `-Xmx1024m` so it will be interesting to see how those play together.

Zapadlo:
Thats interesting...

Should we limit systemd unit files to 1G instead?

I don't know what would happen, as far as service is concerned it will think it is running on a box with 512M ram.

khatton-ft:
If it uses that much memory then I'd guess the JVM will eventually throw OOM errors. Going forward, the ingester will be receiving and handling larger messages (content JSON in payload).

It's worth trying this out to see what happens. If there aren't any problems we should update the Dockerfile to match, otherwise maybe you'll need to increase the memory here.

What if any is the current limit?

Zapadlo:
There isn't, omitting it defaults to `infinity` which is box's max ram, so in our case 8G. Which is why we are getting warnings about low memory as java apps and vulcan are eating several Gs.

khatton-ft:
So in that case :+1: but monitor carefully.

kovacshuni:
I think forcing it from the outside (service file) will not bring good times. It would be much better to limit the java apps' memory by the Xmx flags, that way it will try to stay low by itself. Of course the setting in the service file can remain, but that would be just like a safety net, a strict barrier.

Zapadlo:
Problem is `Xmx` does not actually limit the memory usage of the app to that value.

martingartonft:
should we default to not specifying heap parameters and letting java decide (including deciding based on limits presented by docker)

Zapadlo:
http://git.svc.ft.com/projects/CP/repos/ingester/pull-requests/56/overview

Zapadlo:
Deployed to XP

kovacshuni:
cool stuff

Zapadlo:
Deployed to pre-prod

PR #422, opened 2016-03-16
Upped Content Importer version - now calls the transformer to get the…

martinroddam:
… payload and adds it to the Kafka message on CmsPublicationEvents


khatton-ft:
:+1:

PR #425, opened 2016-03-17
Add platform version to content brands relationships2

scott-ace-newton:
Bump of content-rw-neo4j version which pulls in the below changes

Financial-Times/content-rw-neo4j@9f9f9b2 

-  Added platformVersion property to IS_CLASSIFIED_BY relationships between content and v2 brands.

- content-rw-neo4j returns 204 when content has no body.


Zapadlo:
+1

PR #430, opened 2016-03-21
Re-promote content-importer v45, this time with service changes.

martinroddam:
The change to add the TRANSFORMER_PROXY environment variable was
previously missing from the content-importer service file. This branch
is for updating the content-importer only, so the value is not removed
from the ingesters at this point.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *coverage.*
 *.iml
 .idea/
+/.project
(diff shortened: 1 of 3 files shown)


khatton-ft:
:+1:

Zapadlo:
+1

catalin-ilea:
lgtm

PR #432, opened 2016-03-21
Adding service files for public-organisations-api.

carlosroman:


public-organisations-api@.service
```diff
@@ -0,0 +1,26 @@
...
+# Change killmode from "control-group" to "none" to let Docker remove work correctly.
+KillMode=none
+ExecStartPre=-/bin/bash -c '/usr/bin/docker kill %p-%i > /dev/null 2>&1'
+ExecStartPre=-/bin/bash -c '/usr/bin/docker rm %p-%i > /dev/null 2>&1'
+ExecStartPre=/bin/bash -c '/usr/bin/docker history coco/public-organisations-api:$DOCKER_APP_VERSION > /dev/null 2>&1 || /usr/bin/docker pull coco/public-organisations-api:$DOCKER_APP_VERSION'
+
+ExecStart=/bin/sh -c '\
+  export APP_PORT=8080; \
+  export ENV=$(/usr/bin/etcdctl get /ft/config/environment_tag); \
+  export NEO_HOST_PORT=$(/usr/bin/etcdctl get /ft/services/neo4j/servers/1); \
```
martingartonft: should this go via vulcand, since this is a public read service and we want to lb across neo instances?

martingartonft: or are we achieving something similar in some other way?

catalin-ilea: I was planning on doing that as part of the neo4j resiliency work but if @carlosroman wants to do it that would be great. We basically need to be able to specify a Host header to be passed in when communicating with neo(Host: neo4j-read). I will work on configuring vulcan to LB across neo instances based to that header.

martingartonft: I wonder whether it's too early to think of something other than header based routing now?
If NEO_HOST_PORT was really a baseurl, we could use the double-underscore based service routing that we have in vulcand, meaning no changes to the application.
thoughts?

catalin-ilea: Sure, if we're ok with this long term. That's how I configured routing for publish services(based on double-underscore) but was planning to change that to Host routing later. Yes, no code changes are required in this case.

catalin-ilea: @carlosroman If everyone is ok with double-underscore routing feel free to merge this as is and I'll update it once I have the routing in place.

catalin-ilea: @carlosroman I've configured the routing. Can you please follow the same pattern? https://github.com/Financial-Times/up-service-files/blob/master/public-people-api%40.service#L19

carlosroman: Will update it.

public-organisations-api@.service
```diff
@@ -0,0 +1,34 @@
...
+  /usr/bin/docker run --rm --name %p-%i \
+  -p $APP_PORT \
+  -e "NEO_URL=http://$HOSTNAME:8080/__neo4j-read/db/data" \
```
carlosroman: @catalin-ilea Is this better?

catalin-ilea: lgtm

public-organisations-api@.service
```diff
@@ -0,0 +1,34 @@
...
+  -e "APP_PORT=$APP_PORT" \
+  -e "GRAPHITE_ADDRESS=graphite.ft.com:2003" \
+  -e "GRAPHITE_PREFIX=coco.services.$ENV.public-organisations-api.%i" \
+  -e "LOG_METRICS=false" \
+  -e "CACHE_DURATION=24h" \
```
catalin-ilea: 24h?

carlosroman: 24H is what we run with in production. We update the data only once a week so 24h is more than enough.

catalin-ilea:
lgtm

PR #436, opened 2016-03-23
Upgraded API-policy-component to v25

ppragnya:


Zapadlo:
Heya, we generally don't raise PR's for lower environments to keep the changes and the discussion in one place.

Could you raise a PR for `master` and merge it to `pre-prod` manually instead to deploy it?

Thanks!

PR #446, opened 2016-03-24
Read side support of tagging with locations and genres 

tosan88:
&& fixes in public-annotations-api. See release history: https://github.com/Financial-Times/public-annotations-api/releases


Zapadlo:
+1

tosan88:
This PR should be put ON HOLD, since there is an issue around using public-annotations-api with neo-model-utils library changed to return type hierarchy instead of just direct type...

tosan88:
Testing in pre-prod. Looks good

PR #455, opened 2016-03-28
Making sure neo4j driver always connects to neo4j via vulcan.

catalin-ilea:
This not ideal but it works. 
Neo4j/neo4j driver try to do all sorts of clever things and compute their own urls when they try to connect to neo using the Host header of the first response. This means that the 2nd request to neo4j will not go via vulcan. As a side-effect the app becomes unhealthy when the neo4j instance which the app initially connected to goes down. 
{ "TrustForwardHeader": true, "PassHostHeader": true } solves this problem and routes all the requests via vulcan.

The second problem, now that all requests go via vulcan is that the driver also overwrites the path it uses to connect: /__neo4j-read/db/data becomes /db/data. As a workaround neo4j-read-db was added.

As a conclusion neo4j and its driver don't play nice when you overwrite the Host header or rewriter the url path. We should consider updating the public read apis to set a header(not Host) to be used for routing.

### DIFF
--- a/neo4j-read-blue-sidekick@.service
+++ b/neo4j-read-blue-sidekick@.service
@@ -5,6 +5,7 @@ After=neo4j-blue@%i.service
 
 [Service]
 ExecStart=/bin/sh -c '\
+   etcdctl set /vulcand/backends/neo4j-read/backend \'{"Type": "http"}\'; \
    etcdctl set /vulcand/frontends/neo4j-read/frontend \'{"Type": "http", "BackendId\": "neo4j-read", "Route": "PathRegexp(`/__neo4j-read/.*`)", "Settings": { "TrustForwardHeader": true, "PassHostHeader": true }}\'; \
    etcdctl set /vulcand/frontends/neo4j-read/middlewares/rewrite \'{"Id":"rewrite", "Type":"rewrite", "Priority":1, "Middleware": {"Regexp":"/__neo4j-read(/.*)", "Replacement":"$1"}}\'; \
    etcdctl set /vulcand/frontends/neo4j-read-db/frontend \'{"Type": "http", "BackendId": "neo4j-read", "Route": "PathRegexp(`^/db/data/.*`) && Header(`User-Agent`, `neoism`)", "Settings": { "TrustForwardHeader": true, "PassHostHeader": true }}\'; \
--- a/neo4j-read-blue-sidekick@.service
+++ b/neo4j-read-blue-sidekick@.service
@@ -5,10 +5,10 @@ After=neo4j-blue@%i.service
 
 [Service]
 ExecStart=/bin/sh -c '\
-  etcdctl set /vulcand/backends/neo4j-read/backend \'{"Type": "http"}\'; \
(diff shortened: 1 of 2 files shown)


tamas-molnar:
+1

neo4j-read-blue-sidekick@.service
```diff
@@ -5,10 +5,10 @@ After=neo4j-blue@%i.service
 
 [Service]
 ExecStart=/bin/sh -c '\
-  etcdctl set /vulcand/backends/neo4j-read/backend \'{"Type": "http"}\'; \
```
tosan88: Is this etcd value set deleted on purpose?

Zapadlo:
+1

neo4j-read-blue-sidekick@.service
catalin-ilea: Nope. Good catch. Added it back.

neo4j-read-red-sidekick@.service
```diff
@@ -4,16 +4,17 @@ BindsTo=neo4j-red@%i.service
...
-  etcdctl set /vulcand/backends/neo4j-read/backend '{\"Type\": \"http\"}'; \
-  etcdctl set /vulcand/frontends/neo4j-read/frontend '{\"Type\": \"http\", \"BackendId\": \"neo4j-read\", \"Route\": \"PathRegexp(`/__neo4j-read/.*`)\"}'; \
-  etcdctl set /vulcand/frontends/neo4j-read/middlewares/rewrite '{\"Id\":\"rewrite\", \"Type\":\"rewrite\", \"Priority\":1, \"Middleware\": {\"Regexp\":\"/__neo4j-read(/.*)\", \"Replacement\":\"$1\"}}'; \
+ExecStart=/bin/sh -c '\
+  etcdctl set /vulcand/backends/neo4j-read/backend \'{"Type": "http"}\'; \
+  etcdctl set /vulcand/frontends/neo4j-read/frontend \'{"Type": "http", "BackendId\": "neo4j-read", "Route": "PathRegexp(`/__neo4j-read/.*`)", "Settings": { "TrustForwardHeader": true, "PassHostHeader": true }}\'; \
+  etcdctl set /vulcand/frontends/neo4j-read/middlewares/rewrite \'{"Id":"rewrite", "Type":"rewrite", "Priority":1, "Middleware": {"Regexp":"/__neo4j-read(/.*)", "Replacement":"$1"}}\'; \
+  etcdctl set /vulcand/frontends/neo4j-read-db/frontend \'{"Type": "http", "BackendId": "neo4j-read", "Route": "PathRegexp(`^/db/data/.*`) && Header(`User-Agent`, `neoism`)", "Settings": { "TrustForwardHeader": true, "PassHostHeader": true }}\'; \
```
carlosroman: ```&& Header(`User-Agent`, `neoism`)``` This feels a little bit fragile. It will work for the time being but does mean anything that doesn't use neoism needs to use that header. Something to remember when debuging connection issues.

catalin-ilea: I totally agree with you. We should address this soon.

sjwells: I'm not clear on why the user agent needs to be specified anyway?

catalin-ilea: Because I don't want routing to clash with the one we're doing for Neo4j GUI. The path is the same for those requests as well.

tosan88:
:+1:

PR #457, opened 2016-03-29
Silence sidekick output

Zapadlo:
Currently a cluster will produce ~ 3500 log entries in 10 minutes from
sidekicks alone. Burns network/splunk licence and CPU.

Silences `etcdctl set` output


tosan88:
:+1:

PR #489, opened 2016-04-12
Making sure vulcan backends to not expire for public-annotations-api

catalin-ilea:
We can't get rid of sidekicks yet because of the deployer. @tamas-molnar is updating deployer to be able to cope with missing sidekicks for sequential deployment.

Any thoughts before I implement the same chaneg for the rest of our read apis?

### DIFF
--- a/public-annotations-api-sidekick@.service
+++ b/public-annotations-api-sidekick@.service
@@ -17,6 +17,7 @@ ExecStart=/bin/sh -c "\
   done; \
   etcdctl set /ft/services/$SERVICE/servers/%i http://%H:$PORT;"
 
+ExecStop=/usr/bin/etcdctl rm /ft/services/public-annotations-api/servers/%i
 Restart=on-failure
 RestartSec=60
 
--- a/public-annotations-api-sidekick@.service
+++ b/public-annotations-api-sidekick@.service
@@ -6,15 +6,16 @@ After=public-annotations-api@%i.service
 [Service]
 RemainAfterExit=yes
 ExecStart=/bin/sh -c "\
-  etcdctl set   /ft/services/public-annotations-api/healthcheck     true; \
-  etcdctl mkdir /ft/services/public-annotations-api/servers; \
-  etcdctl set /ft/healthcheck/public-annotations-api-%i/path /__health; \
+  export SERVICE=$(echo %p | sed "s/-sidekick//g"); \
+  etcdctl set   /ft/services/$SERVICE/healthcheck     true; \
+  etcdctl mkdir /ft/services/$SERVICE/servers; \
+  etcdctl set /ft/healthcheck/$SERVICE-%i/path /__health; \
   while [ -z $PORT ]; do \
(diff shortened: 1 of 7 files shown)


Zapadlo:
Could you copy this pattern for service name? https://github.com/Financial-Times/up-service-files/blob/master/public-concordances-api-sidekick%40.service#L8

tamas-molnar:
lgtm

catalin-ilea:
@Zapadlo Done

public-annotations-api-sidekick@.service
```diff
@@ -4,17 +4,19 @@ BindsTo=public-annotations-api@%i.service
...
+  etcdctl set /ft/healthcheck/$SERVICE-%i/path /__health; \
+  while [ -z $PORT ]; do \
+    PORT=`echo $(/usr/bin/docker port $SERVICE-%i 8080 | cut -d':' -f2)`; \
+    echo 'Port not set, retrying'; \
     sleep 5; \
-  done"
+  done; \
+  etcdctl set /ft/services/$SERVICE/servers/%i http://%H:$PORT;"
 
-ExecStop=/usr/bin/etcdctl rm /ft/services/public-annotations-api/servers/%i
```
Zapadlo: Why do we not remove it anymore?

tosan88:
:+1:

public-annotations-api-sidekick@.service
catalin-ilea: Because I'd rather have the sidekicks re-write this key when they are restarted rather than deleting it and then setting it. This is a workaround for the deployer issue(deployer randomly restarting sidekicks).

catalin-ilea: Can you think of any reason why we should remove it?

Zapadlo: If remove the service or reduce number of instances. Not that frequent, otherwise we can end up with bad config that will never get fixed.

Deployer should get fixed seperately, shouldn't work around deployer issues here

catalin-ilea: Ok, I'll add it back.

catalin-ilea: Just an fyi we currently need to manually clean-up config when we remove services.

Zapadlo: What config? VCB is supposed to unset values it created after `/ft/services` values are gone..

kovacshuni:
+1

public-annotations-api-sidekick@.service
catalin-ilea: VCB cleans up, yes, but nothing deletes /ft/services/people-rw-neo4j-red for example. We only delete /ft/services/people-rw-neo4j-red/servers/%i

Zapadlo: Whatever sets that value should be responsible for removing it. Anyway... +1

Zapadlo:
+1

PR #493, opened 2016-04-14
Bumped public organisation api version from v0.0.41 -> v0.0.43

scott-ace-newton:


nwrigh:
:+1: As long as you realise that you are pulling in Sarah's change to neo model utils as well but your testing should prove no regression

PR #505, opened 2016-04-20
:Adding timeouts to public backends.

carlosroman:
Also added the same to our neo4j sidekicks.

### DIFF
--- a/public-people-api-sidekick@.service
+++ b/public-people-api-sidekick@.service
@@ -11,7 +11,7 @@ ExecStart=/bin/sh -c "\
   etcdctl mkdir /ft/services/$SERVICE/servers; \
   etcdctl set /ft/healthcheck/$SERVICE-%i/path /__health; \
   etcdctl set /ft/healthcheck/$SERVICE-%i/categories read; \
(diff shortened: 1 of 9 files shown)


catalin-ilea:
Could you please add the same for the rest of the public apis and api-component-policy?

Zapadlo:
We don't have a default value in vcb then?

+1

catalin-ilea:
We should have this in VCB from my point of view. VCB supports adding public frontends but I guess we forgot about that when we implemented the pattern in the sidekicks

public-people-api-sidekick@.service
```diff
@@ -11,6 +11,7 @@ ExecStart=/bin/sh -c "\
   etcdctl mkdir /ft/services/$SERVICE/servers; \
   etcdctl set /ft/healthcheck/$SERVICE-%i/path /__health; \
   etcdctl set /ft/healthcheck/$SERVICE-%i/categories read; \
+  etcdctl set /vulcand/backends/public-services-people/backend \'{\"Type\": \"http\", \"Settings\": {\"KeepAlive\": {\"MaxIdleConnsPerHost\": 256, \"Period\": \"35s\"}}}\'; \
```
catalin-ilea: I don't think this does what is supposed to do. These settings should be applied to backend vcb-public-people-api

catalin-ilea: We need to clean up these sidekicks and configure this in vulcan-config-builder

PR #506, opened 2016-04-21
HA for aggregator-healthcheck

Zapadlo:
Now a system critical service we want to ensure high % uptime

### DIFF
--- a/services.yaml
+++ b/services.yaml
@@ -3,7 +3,7 @@ services:
   count: 2
 - name: aggregate-healthcheck@.service
   version: v1.0.7
-  count: 1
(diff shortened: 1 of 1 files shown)


tosan88:
:+1:

martingartonft:
This will double our application load on the __heath endpoints across the cluster.  Do we care?  Should we reduce the refresh rate to compensate?
In future we could use an etcd lock to ensure that multiple HC instances are not calling all the services at the same time.

Zapadlo:
Funny, I said exactly the same, its a concern, soaking it `xp` right now.

PR #518, opened 2016-04-29
pam: skip notification check 4 deletes if there is no previous notification for that content

Nesze:
Includes @JuliaFernee's changes as well (0.17.1).
Both have been tested in lower environments (xp, dynpub).

Will be in pre-prod for the weekend.


JuliaFernee:
+1
I've deployed 0.17.1 to prod clusters

PR #522, opened 2016-05-10
Adding prod bridges

tamas-molnar:


cms-kafka-bridge-prod@.service
```diff
@@ -0,0 +1,28 @@
...
+# Change killmode from "control-group" to "none" to let Docker remove work correctly.
+KillMode=none
+ExecStartPre=-/bin/bash -c '/usr/bin/docker kill %p-%i > /dev/null 2>&1'
+ExecStartPre=-/bin/bash -c '/usr/bin/docker rm %p-%i > /dev/null 2>&1'
+ExecStartPre=/bin/bash -c 'docker pull coco/coco-kafka-bridge:$DOCKER_APP_VERSION'
+
+ExecStart=/bin/sh -c '\
+  export APP_PORT=8080; \
+  export QUEUE_PROXY_ADDRS=https://kafka-proxy-iw-uk-p-1.glb.ft.com,https://kafka-proxy-iw-uk-p-2.glb.ft.com; \
+  export GROUP_ID=$(/usr/bin/etcdctl get /ft/config/environment_tag)"-kafka-bridge-prod"; \
```
BerniVarga: Couldn't be used the same group_id we used before? without the prod tag at the end...

tosan88:
:+1:

cms-kafka-bridge-prod@.service
tamas-molnar: Yeah, but this is also unique, so it doesn't really matter. It was useless to change, but it'll work.

PR #526, opened 2016-05-12
Adding prod bridges

tamas-molnar:
- same changes as in XP, made separate branches as these are branch-specific and won't be merged back to master


tosan88:
:+1:

PR #539, opened 2016-05-19
mat reject blank body

galiri:
change to prevent publish of blank body tags post transformation e.g. <body>  \n\n <body>


khatton-ft:
:+1:

PR #540, opened 2016-05-19
Feature/notificationrw create indices

peteclark-ft:
Notification RW service will now ensure the mongo notification collection indexes are created during its startup routine. If the indexes already exist, this should have no impact.


yolandeleungFT:
+1

PR #542, opened 2016-05-19
Upgrade document store to support concept and type on lists

sjwells:
And a new query method to return the list based on type and concept


khatton-ft:
:+1:

PR #545, opened 2016-05-23
Bump content-importer and WPIM versions.

khatton-ft:
These are the two services that produce messages for the
CmsPublicationEvents topic. (There is no requirement to deploy them
together, but it is convenient).


izzyblues:
+1

PR #546, opened 2016-05-23
Fix mongodb configuration

martingartonft:
Removed mongodb-all and update mongodb configurator to wait for exactly
3 instances before running.


martingartonft:
btw, I tested this a few of times in new clusters and it seems fine (after about 1000 attempts at getting it right)

PR #557, opened 2016-05-27
Added bridges for UCS Test. These should be deployed in lower envs only.

catalin-ilea:
With this change lower env clusters will automatically consume traffic from both UCS Test and Prod without any manual changes.
@yolandeleungFT @nwrigh please cherry-pick this onto team branches
@nwrigh @scott-ace-newton  You might want to follow the same pattern for the bridges you guys are about to add


catalin-ilea:
Tested in XP

Zapadlo:
Is this the first change that permanently diverges branches?

catalin-ilea:
Nope...we have had different number of services deployed in lower envs compared to prod for some time now.

Nesze:
lgtm
(apart from the order of services and sidekick, but that's the formatter's fault)

PR #561, opened 2016-05-31
methode-article transformer to v 60 allow publish of EOM Story with i…

galiri:
MAT allows publish of invalid workflow status before enforce date (1 June 2011)


khatton-ft:
:+1:

PR #567, opened 2016-06-06
Bump api-policy-component version.

khatton-ft:
Intermediate builds were related to FTPlatform migration only.


catalin-ilea:
+1

PR #570, opened 2016-06-07
Videos in articles

kovacshuni:
tested in XP
pre-prod is currently not fit for testing this.


kovacshuni:
tested in pre-prod

BerniVarga:
+1

PR #571, opened 2016-06-08
Six degrees app

tomaszlibich:


public-six-degrees-app.service
```diff
@@ -0,0 +1,31 @@
...
+# let Docker remove work correctly.
+KillMode=none
+
+# stop already running instance
+ExecStartPre=-/usr/bin/docker kill %p
+
+# remove existing instance
+ExecStartPre=-/usr/bin/docker rm %p
+
+ExecStart=/bin/sh -c "/usr/bin/docker run --name %p;"
```
Zapadlo: 1. This will not expose any ports, if you used `EXPOSE` in `Dockerfile` - add `-P`, if not add `-p <container-port>
2. Add `--rm` to remove the container after it stops
3. Filename is missing `@`
4. `--name` convention is `%p-%i`
5. You need to specify the docker image to run, format is: `docker.io/coco/six-degrees:0.0.1 # repo/author/image/version`

tomaszlibich: Changes applied as per your suggestions. Thanks.

martingartonft:
This will depend on elastic being merged into semantic.  I''ll create a PR for that now.

PR #572, opened 2016-06-08
Topics-rw - adding identifier nodes

BerniVarga:
Changes include for the topics-rw-neo4j:
* Make all the alternative ids Identifier nodes
* Only use prefLabel
* Add build-info
* Update the RW for the new json model - coming out from the transformer


BerniVarga:
Tested in xp.

tosan88:
:+1:

PR #583, opened 2016-06-10
Adding genre transformer

nwrigh:


tosan88:
Strange that genres-transformer wasn't added yet. Anyway
:+1:

PR #588, opened 2016-06-13
Identifiers in sections transformer

nwrigh:
* Add build-info
* Use prefLabel and remove canonical name
* All identifiers moved to using the new approach (see example above)
* Add healthcheck
* Connect to TME directly not via structure service


tosan88:
:+1:

PR #589, opened 2016-06-13
Bumped version of topics transformer to use new Taxonomy name and remove LEI and Factset references

nwrigh:


BerniVarga:
+1

tosan88:
The change in the `subjects-transformer@.service` file is already done in another branch: https://github.com/Financial-Times/up-service-files/pull/590/commits/ceb524356930519bd547bda7469e39b66391fc4f, so the commit on this branch will cause a conflict. Can you please revert it?

PR #593, opened 2016-06-14
Bumped the version for special reports transformer

nwrigh:


tosan88:
:+1:

martingartonft:
I can see all the version bumps and they look fine, but what's all the new services that have different environment names in them?   foo-pre-prod-bar?

PR #598, opened 2016-06-14
Added binding-service-specific v1-content-annotators

tamas-molnar:
* needed to process the messages coming from the binding-service, see UPP-1262
* only change from the existing v1-content annotators is the `-env "ANNOTATING_SYSTEM=http://cmdb.ft.com/systems/binding-service"`
* tested in XP, works
* these will be removed when the existing V1-metadata publish flow will be decommed (the one which goes through preditor)


martingartonft:
+1

tosan88:
The kafka group ids should be different than for existing services...

tamas-molnar:
Good spot @tosan88, changed!

tosan88:
:+1:

PR #601, opened 2016-06-14
Bumped the version of only the special reports transformer

nwrigh:


BerniVarga:
+1

tosan88:
Just for others: the branch name is misleading, this is for special-reports-transformer :)
:+1:

PR #609, opened 2016-06-15
Bumping api policy component to v40

neetuFT:
API policy component to add validatelinkedresources=true parameter


yolandeleungFT:
+1

PR #613, opened 2016-06-21
bump CPR - expose standout section

Nesze:
fix: http://git.svc.ft.com/projects/CP/repos/content-public-read/pull-requests/17/overview

Tested in xp & pre-prod.


khatton-ft:
:+1:

catalin-ilea:
A bit late to the party...but enable sequential deployment on content-public-read@.service? :)

PR #614, opened 2016-06-21
Ack support for agg-hc

tosan88:
+ adding `sequentialDeployment` flag

**Note**: there is no GUI support to ACK a service, you have to do it manually, as described in https://github.com/Financial-Times/aggregate-healthcheck/blob/master/README.md


catalin-ilea:
lgtm

PR #617, opened 2016-06-22
Whitelist neo4j backup

tamas-molnar:
* tested in `semantic`, works as excepted
* merged to `xp` too, but since neo4j-backup isn't there it did nothing
* will only push to `pre-prod` for now, `prod` after the ice age


tosan88:
:+1:

kovacshuni:
+1

PR #619, opened 2016-06-29
Store video notifications

Nesze:
Do not merge yet. A few service versions still have to be incremented.

Tested in xp.


khatton-ft:
lgtm but see my comments in notifications-rw.

Nesze:
Tested in pre-prod.

PR #620, opened 2016-06-30
Public-Brands-Api - Adjust to new identifier structure + containerisation

scott-ace-newton:
Added service and sidekick files. added public-brands-api entry to services.yaml


khatton-ft:
:+1:

tosan88:
:+1:

PR #621, opened 2016-06-30
concepts bridges - enable autocommit, set instance number to 1

tamas-molnar:
* I made a branch for each env since prod doesn't have pre-prod bridges, for example, and some bridges were updates manually straight on the env branch
* tried this in lower envs
* you need to destroy the number 2 nodes manually after the merge


scott-ace-newton:
:+1:

martingartonft:
+1

PR #622, opened 2016-06-30
concepts bridge - enable autocommit, set instance number to 1

tamas-molnar:
same changes as this: https://github.com/Financial-Times/up-service-files/pull/621 for prod


tosan88:
:+1:

PR #625, opened 2016-07-04
Reduce metadata transformers count

tosan88:
One instance per transformer should be enough, thus reducing memory/CPU usage and load on TME on startup.

Note: deployer doesn't pick up changes on counts.


catalin-ilea:
+1

martingartonft:
+1

PR #626, opened 2016-07-04
Remove all v1 relationships at delete (not only MENTIONS relationships)

BerniVarga:
Tested in xp and pre-prod


tosan88:
:+1:

catalin-ilea:
+1

PR #637, opened 2016-07-06
Feature/doc store api concept listtype index

izzyblues:
I added the creation of a index on lists at startup.
The index improves performances on queries by listType and Concept properties.
Such queries are used by Next to render lists in section pages.


khatton-ft:
:+1:

tosan88:
That's quite a big version change, but I see that additional changes are related to some scripts within the app ,which should be fine. So
:+1:

PR #639, opened 2016-07-06
Updated the version of people-rw-neo4j for updating the data model to…

ppragnya:
… add new fields with old identifiers


izzyblues:
👍

khatton-ft:
:+1:

PR #642, opened 2016-07-08
Add identifiers organisation-rw

nwrigh:
Update organisations-rw to v0.1.4
https://github.com/Financial-Times/organisations-rw-neo4j/releases/tag/0.1.4

The change is so the new identifiers are written but also concordance will work


BerniVarga:
+1

nwrigh:
Released in another PR

PR #645, opened 2016-07-11
Added single instance elasticsearch

martingartonft:


services.yaml
```diff
@@ -116,6 +116,8 @@ services:
   version: v76
   count: 2
   sequentialDeployment: true
+- name: elasticsearch-sidekick.service
+- name: elasticsearch.service
```
duffj: No version number?

elasticsearch.service
```diff
@@ -0,0 +1,23 @@
...
+[Service]
+TimeoutStartSec=0
+# Change killmode from "control-group" to "none" to let Docker remove
+# work correctly.
+KillMode=none
+ExecStartPre=-/bin/sh -c 'docker kill %p > /dev/null 2>&1'
+ExecStartPre=-/bin/sh -c 'docker rm %p > /dev/null 2>&1'
+ExecStartPre=/usr/bin/docker pull elasticsearch:2.3.3
+ExecStart=/bin/sh -c "\
+  /usr/bin/docker run --rm --name %p -v /vol/elasticsearch:/usr/share/elasticsearch/data -P elasticsearch:2.3.3 -Dnetwork.host=0.0.0.0"
```
duffj: Latest version is 2.3.4 (https://hub.docker.com/_/elasticsearch/), why use 2.3.3?

elasticsearch.service
```diff
@@ -0,0 +1,23 @@
...
+ExecStop=/usr/bin/docker stop -t 3 %p
+Restart=on-failure
+RestartSec=60
+
+[X-Fleet]
+Conflicts=%p.service
+MachineMetadata=persistent_tag=3
```
duffj: There is quite a lot running on machine 3 now (presumably because it's the one with the big EBS volume mounted); should we be concerned about it?

services.yaml
martingartonft: The version is fixed in the service file.  I believe we have this approach elsewhere for other third party apps that we update infrequently.

elasticsearch.service
martingartonft: I can easily move it.  Which might be better do you think?

nwrigh:
No longer relevant as the ES work is a whole stream to complete in Q4

PR #656, opened 2016-07-13
Feature/enable neo4j http logging

duffj:


khatton-ft:
only for red?

duffj:
Yep, just for now, until I can get it working.

duffj:
The PR for `master` will be for red and blue.

khatton-ft:
Makes sense. I hadn't spotted this PR wasn't for master.
:+1:

duffj:
@khatton-ft thanks. I should make it clear that this is a totally temporary solution taken from https://neo4j.com/developer/docker-23/ which allows us to override specific config items. It's not actually production-ready at all. We'll have to come up with a significantly better (more automated and more surgical) way of overriding config to enable request logging before it's ready to deploy to prod. So this is more of a prototyping solution to see how helpful request logging is.

And actually, with the problem we're currently looking at, it hasn't been needed (so far).

tosan88:
:+1:

nwrigh:
We now use config files for configuration of Neo not this method

PR #659, opened 2016-07-14
Api host is now cluster specific for notifications-rw

catalin-ilea:
Api host is now cluster specific for notifications-rw
Notifications-rw v48 - encoding url params for links href


khatton-ft:
:+1:

martingartonft:
+1

Nesze:
https://xp-up.ft.com/content/notifications?since=2016-07-16T00:00:00.000Z

This link `http://xp-up-read.ft.com/content/notifications?since=2016-07-16T00%3A00%3A00Z` is not correct, wrong protocol. Beside, do we want to return to b2b clients coco urls, potentially them, trying to bypass the gateway?

catalin-ilea:
@Nesze In pre-prod and prod this link will point to Mashery which support http. COCO urls will not reach B2B clients. As for the protocol, historically, all the links we expose are http, maybe we can migrate to https but that needs more coordination as it might impact our clients. Yes, I agree, the link won't work in team clusters...do we want to change the code to fix this(I suspect this is not the only service that would require changing)?

Nesze:
No, it's fine then. I was worried only about prod/pre-prod links. 

👍

tosan88:
:+1:

PR #667, opened 2016-07-20
Reduced instances of orgs, people, and membership rw to try to limit …

scott-ace-newton:
…the number of neo4j deadlock errors we see whilst mass ingesting concepts


martingartonft:
+1

catalin-ilea:
+1

PR #669, opened 2016-07-20
Bump the version of roles-rw-neo4j

sjwells:


tosan88:
+1 but please coordinate with the work around concordances before merging it

nwrigh:
Any chance you can reduce the number of instances to one? This will avoid any deadlocks

PR #678, opened 2016-07-25
Concept bridges

catalin-ilea:
The model we used to far to replicate prod traffic to lower environments doesn't really make sense for concept ingestion pipeline. This also puts unnecessary load on the pub-prod kafka proxy for no good reason and impacts the overall performance of the ingestion.


BerniVarga:
+1

PR #687, opened 2016-08-02
Updated the version of the org-rw to be the final version

nwrigh:
We had to release a temporary version of the org-rw that went through and handled all the parent organisations first. However this is the final version .

https://github.com/Financial-Times/organisations-rw-neo4j/releases/tag/0.2.0


BerniVarga:
+1

kovacshuni:
+1

PR #689, opened 2016-08-03
Unique container names ingestion flow

BerniVarga:
Fixing "container already in use" and "port not being set" issues for ingestion flow.
(See https://jira.ft.com/browse/UPP-1388 and https://jira.ft.com/browse/UPP-1393 for more details.)

Some service file refactoring was also made.

### DIFF
--- a/brands-rw-neo4j-blue-sidekick@.service
+++ b/brands-rw-neo4j-blue-sidekick@.service
@@ -11,13 +11,13 @@ ExecStart=/bin/sh -c "\
   etcdctl mkdir /ft/services/$SERVICE/servers; \
   etcdctl set /ft/healthcheck/$SERVICE-%i/path /__health; \
   while [ -z $PORT ]; do \
-    PORT=`echo $(/usr/bin/docker port $SERVICE-%i 8080 | cut -d':' -f2)`; \
-    echo 'Port not set, retrying'; \
     sleep 5; \
+    CONTAINER_NAME=$(docker ps -q --filter=name=\"$SERVICE\"-%i_); \
+    PORT=`echo $(/usr/bin/docker port $CONTAINER_NAME 8080 | cut -d':' -f2)`; \
   done; \
   etcdctl set /ft/services/$SERVICE/servers/%i http://%H:$PORT;"
 
-ExecStop=/usr/bin/etcdctl rm /ft/services/brands-rw-neo4j-blue/servers/%i
+ExecStop=-/bin/bash -c 'etcdctl rm /ft/services/brands-rw-neo4j-blue/servers/%i'
 Restart=on-failure
 RestartSec=60
 
--- a/concepts-kafka-bridge-pub-prod@.service
+++ b/concepts-kafka-bridge-pub-prod@.service
@@ -9,17 +9,17 @@ Environment="DOCKER_APP_VERSION=latest"
 TimeoutStartSec=0
 # Change killmode from "control-group" to "none" to let Docker remove work correctly.
 KillMode=none
-ExecStartPre=-/bin/bash -c '/usr/bin/docker kill %p-%i > /dev/null 2>&1'
-ExecStartPre=-/bin/bash -c '/usr/bin/docker rm %p-%i > /dev/null 2>&1'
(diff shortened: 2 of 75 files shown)


genres-rw-neo4j-blue@.service
```diff
@@ -1,23 +1,29 @@
 [Unit]
 Description=Genres neo4j rw blue service
-After=docker.service
+After=vulcan.service
 Requires=docker.service
-Wants=neo4j-blue@1.service genres-rw-neo4j-blue-sidekick@%i.service
+Wants=genres-rw-neo4j-blue-sidekick@%i.service
```
tosan88: Is `neo4j-blue@1.service` not needed here anymore?

brands-rw-neo4j-blue-sidekick@.service
```diff
@@ -11,13 +11,13 @@ ExecStart=/bin/sh -c "\
   etcdctl mkdir /ft/services/$SERVICE/servers; \
   etcdctl set /ft/healthcheck/$SERVICE-%i/path /__health; \
   while [ -z $PORT ]; do \
-    PORT=`echo $(/usr/bin/docker port $SERVICE-%i 8080 | cut -d':' -f2)`; \
-    echo 'Port not set, retrying'; \
     sleep 5; \
+    CONTAINER_NAME=$(docker ps -q --filter=name=\"$SERVICE\"-%i_); \
```
peteclark-ft: Could this return more than one result, given that we're adding a UUID on the end (and so $SERVICE-%i_ is not necessarily unique)?

concepts-kafka-bridge-pub-prod@.service
```diff
@@ -9,17 +9,28 @@ Environment="DOCKER_APP_VERSION=latest"
 TimeoutStartSec=0
 # Change killmode from "control-group" to "none" to let Docker remove work correctly.
 KillMode=none
-ExecStartPre=-/bin/bash -c '/usr/bin/docker kill %p-%i > /dev/null 2>&1'
-ExecStartPre=-/bin/bash -c '/usr/bin/docker rm %p-%i > /dev/null 2>&1'
-ExecStartPre=/bin/bash -c 'docker pull coco/coco-kafka-bridge:$DOCKER_APP_VERSION'
+ExecStartPre=-/bin/bash -c '/usr/bin/docker kill "$(docker ps -q --filter=name=%p-%i_)" > /dev/null 2>&1'
+ExecStartPre=-/bin/bash -c '/usr/bin/docker rm "$(docker ps -q --filter=name=%p-%i_)" > /dev/null 2>&1'
+ExecStartPre=/bin/bash -c 'docker history coco/coco-kafka-bridge:$DOCKER_APP_VERSION > /dev/null 2>&1 || docker pull coco/coco-kafka-bridge:$DOCKER_APP_VERSION'
```
peteclark-ft: Will `docker pull` not cache the image?

peteclark-ft:
Other than the comments - 👍

genres-rw-neo4j-blue@.service
BerniVarga: We should have a dependency between the writers and the underlaying persistence storage. We will do this in a separate story. For know, i removed this additional dependencies for consistency (also, we are not sure that this is the way how we should define them).

I also modified the 2 files you mentioned.

brands-rw-neo4j-blue-sidekick@.service
BerniVarga: It could only happen if we have left-over containers - which shouldn't be the case if the stop command works as expected. Also, with this method of filtering, we shouldn't use "_" in service file names (which we don't do). => $SERVICE-%i_ will always be unique

concepts-kafka-bridge-pub-prod@.service
BerniVarga: Docker pull won't download the image (and underlaying images) if they are already stored. I think it has to go to the public repo, though. (I might be proven wrong...) 
We tried to avoid this by checking if we already have it locally.

tosan88:
:+1:

kovacshuni:
👍

PR #690, opened 2016-08-03
Bump WPIM version.

khatton-ft:
Switch to passive consumer healthcheck; don't throw InterruptedException on normal shutdown.


yolandeleungFT:
+1

PR #693, opened 2016-08-04
Service file template

BerniVarga:
An example service file with it's sidekick - with some explanation inside.

I can add more comments, i can remove them, or we can just add a link in the README to the service-file which we consider a "role model" instead of keeping a template up-to-date.

I would like to hear your thoughts about this. :)


kovacshuni:
👍

tosan88:
:+1:

PR #696, opened 2016-08-04
Fixed to the right version v0.2.0 of org-rw as mistakedly got removed

nwrigh:
During a merge conflict the wrong version ended up in the org-rw (should be 0.2.0)

(Previous PR: https://github.com/Financial-Times/up-service-files/pull/687)


scott-ace-newton:
:+1:

PR #698, opened 2016-08-05
Fix version of mongo-backup service.

khatton-ft:
Fix the version number so that later merges for the service aren't picked up until desired.
The most recent Jenkins build was v13 so I tagged the Docker image as such (I am not sure why it wasn't already tagged, in Jenkins it looks like it should be). The intended effect of this PR is in fact "no change".


martingartonft:
+1

catalin-ilea:
+1

tosan88:
:+1:

PR #699, opened 2016-08-05
Add unique container names for global units

tosan88:
*Mental note*: This will cause outage, so deploy it per datacenter!


BerniVarga:
+1

kovacshuni:
+1

PR #705, opened 2016-08-09
Vulcan config builder v1.3.0 

catalin-ilea:
No failover predicate by default which means no retries on error

https://github.com/Financial-Times/vulcan-config-builder/pull/17


BerniVarga:
+1

tosan88:
:+1:

PR #713, opened 2016-08-16
Mongo backup service run on all mongodb nodes

khatton-ft:
Run a mongo-backup service on each mongodb node. The updated service version shuts down unless it is running on the lowest secondary node (in lexocographic order), so a backup should always be made (unless there is no secondary). Tested in dynpub.


martingartonft:
+1

tosan88:
:+1:

kovacshuni:
+1

khatton-ft:
Note: It seems that although the deployer removes the un-templated timer, it doesn't tidy up the un-templated mongodb.service, which has to be destroyed manually.

tosan88:
Also it seems even if the desiredState is set to be loaded, the apps look inactive/dead :/

tosan88:
:+1: 
but note certain failures could leave the service `inactive` instead of `failed`

yolandeleungFT:
+1

PR #722, opened 2016-08-22
Update logfilter

martingartonft:
Update logfilter to a version that can extract response times from some
of the logs it parses.


nwrigh:
:+1:

PR #723, opened 2016-08-22
Use ft neo4j

martingartonft:
Use ft custom neo4j docker image, which is based on alpine linux, has configurable request logging, and also set up new memory limits, including a cgroup based limit at the container level.


nwrigh:
:+1:

kovacshuni:
+1

PR #727, opened 2016-08-23
Decouple annotations API errors from enriched content API

david-smith-ft:
Annotation API errors will no longer cause the Enriched Content API to
error, it will just return an empty array


nwrigh:
:+1:

PR #728, opened 2016-08-23
Fleet unit hc whitelist regex

khatton-ft:
Allows the fleet-unit-healthcheck to whitelist `mongo-backup@.service`. Tested in dynpub,


izzyblues:
👍

PR #729, opened 2016-08-24
Revert fleet-unit-healthcheck version.

khatton-ft:
Switch back in prod to the version of fleet-unit-healthcheck that will be healthy for the singleton mongo-backup service.


yolandeleungFT:
+1

PR #730, opened 2016-08-24
Updated dns registrators. 

catalin-ilea:
Making sure we don't remove records as part of registering elb and tunnel records.

https://github.com/Financial-Times/coco-elb-dns-registrator/pull/5


catalin-ilea:
Tested in XP

BerniVarga:
+1

PR #745, opened 2016-08-31
Lower resource usage by mongodb backup

martingartonft:
Use newer version which uses newer Go version and had simplified code,
both of which reduce CPU overhead.

Also, limit memory to 1g in the docker command.

### DIFF
--- a/mongo-backup.service
+++ b/mongo-backup.service
@@ -38,8 +38,9 @@ ExecStart=/bin/sh -c "\
         -e DATA_FOLDER=$DATA_FOLDER \
         -e S3_DOMAIN=$S3_DOMAIN \
         -e ENV_TAG=$ENV_TAG \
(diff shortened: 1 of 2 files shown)


tosan88:
:+1:

khatton-ft:
Latest version built by Jenkins was v19!

martingartonft:
I wonder why 13 was the one in use?

Anyway, happy to re-release as v20.0.0 and update here if we think it matters?

khatton-ft:
v18 is in use in Dynpub, but hasn't yet made it to production. I believe you created v19 :)

martingartonft:
I didn't realise the dynpub stuff was not in prod yet, sorry.

I'll hold off deploying this anywhere else and re-version this once dynpub stuff is in prod. I may have more changes by then anyway.

PR #749, opened 2016-09-01
Fleet unit hc mongo backup

khatton-ft:
- Deploy fleet-unit-healthcheck 1.0.7 to allow whitelisting of multiple mongo-backup service instances.
- Remove "type=oneshot" from mongo-backup so that it is listed as "active" rather than "activating" when running, so that normal operation doesn't trigger a warning in the fleet-unit-healthcheck.


tamas-molnar:
+1

PR #750, opened 2016-09-02
Pin org r/w apps to corresponding neo instances.

martingartonft:
Put org r/w apps to the same machines as their neo4j instances.
Whether to put these on the same machine as neo4j is a tradeoff.
Using a different machine reduces resource contention, and using the
same machine reduces inter-vm networking.
I have ovserved that the ingests go slowly when on different vms, and
putting them on the same machine makes things significantly faster.
In practice, resource contention on the neo4j machine is mininal because
the org r/w app uses very little memory of CPU.


nwrigh:
:+1:

scott-ace-newton:
:+1:

tosan88:
:+1:

PR #756, opened 2016-09-07
video-mapper v1.2.9 - HTML escape body.

kovacshuni:
goes together with https://github.com/Financial-Times/pub-service-files/pull/37


khatton-ft:
I don't think that referenced PR is the correct one, but :+1: anyway

tamas-molnar:
+1

PR #767, opened 2016-09-15
Initial version of mongodb online backup

martingartonft:


kovacshuni:
+1

khatton-ft:
Have you checked how this performs when there's a real volume of r/w load going on?
I would suggest scheduling initially at a different time to the existing backup.

martingartonft:
@khatton-ft  I haven't checked that, hence this not being in production yet. :-)
It is already scheduled at a different time to the existing one.

Actually I am tempted to remove the timer from the first release of this so it can go to prod but will only run when manually kicked.

khatton-ft:
Yes, I was a bit confused because 11:00 BST == 12:00.
My experience is that the current backup takes around 90 minutes.
In any case, removing the timer and running on demand sounds like a good way to try this out in prod.

khatton-ft:
:+1:

mongodb-online-backup.service
```diff
@@ -0,0 +1,39 @@
...
+        -e S3_DIR=$S3_DIR \
+        -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID \
+        -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY \
+        --memory="1g" \
+        coco/up-mongolizer:$DOCKER_APP_VERSION /up-mongolizer backup upp-store/content,upp-store/lists,upp-store/notifications; "
+
+[Install]
+# this sets user-level
+WantedBy=multi-user.target
+
```
tosan88: I see the service is not pinned to a mongo machine. Was that for network optimization? https://github.com/Financial-Times/up-service-files/blob/master/mongo-backup%40.service#L48

martingartonft: I didn't want to pin to a single machine. I guess it could be pinned to any persistent. However, since it uses a fair bit of CPU, I don't know if it might be good to keep it away from mongodb?

PR #768, opened 2016-09-16
Allow neo4j some more memory

martingartonft:
We have the JVM heap set to 2G and cache set to 2G, and hard container
limit set to 4G. This leaves no overhead for the JVM or anything else.


kovacshuni:
+1

PR #775, opened 2016-09-20
bridge added for pub-dynpub-uk, other bridges cleaned up

galiri:
This is for Dunpub delivery cluster only.

note authorization key is currently entered manually tot he cluster, until a better solution is in place


yolandeleungFT:
+1

khatton-ft:
I'm a bit concerned about the cases where Git is showing files as renamed. Hopefully it will be fine but I wonder whether this might cause merge conflicts when some later branch is merged into dynpub.
:+1: anyway, because I am not certain there will be any problem.

PR #784, opened 2016-09-23
Updated Neo4J to 2.3.7 - BLUE ONLY

nwrigh:
Updating the blue Neo4J to 2.3.7. We have split them into two PR's so that we don't need to do a ZDD


tosan88:
Kind of confusing that one needs unit specifier ('m') the other doesn't... 
Anyway,
:+1:

PR #796, opened 2016-09-27
Feature/notifications cache ttl

neetuFT:
Feature/notifications cache ttl


neetuFT:
After the review, we will merge into master locally

PR #803, opened 2016-09-28
Set Xms & Xmx for Wordpress Article Transformer

carlosroman:
Load test works comfortably at 169 Req/s with no 502s or 504s. 99th pct is 248 ms.


martingartonft:
Do lower values than 384 show problems?

carlosroman:
Values lower than 384 lead to the app generating 502s or 504s. Will post in slack the graphs.

PR #806, opened 2016-09-28
Removing UCS bridges from Pre-Prod

peteclark-ft:
Manual cleanup of etcd will be performed before/after merge.


galiri:
Please be aware of the manual clean up to etcd keys and  service healthcheck directories.

:+1:

PR #809, opened 2016-09-28
Cpr memory constarints

JuliaFernee:


tosan88:
It looks good, but I'm thinking now we should also hard limit memory by setting memory limit on the docker container itself... 
@MartinGarton @JuliaFernee @khatton-ft @sjwells what are your thoughts? ^^

khatton-ft:
I have no objections to @tosan88's suggestion.
Are there any disadvantages?
What are the risks of _not_ doing that?

martingartonft:
ISTR a conversation about doing that in a separate phase, in part because we haven't gained a solid understanding of how much RES memory the apps will use under real load and with this first phase of config changes in.

The risk of not doing that is arguably low, as long as these jvm settings are doing there job.

The risk of doing it right now is we get the value wrong and things crash and we have to suddenly do the work at short notice to decide a better value.

tosan88:
fair point, @MartinGarton That should have been only a safety net anyway...

PR #818, opened 2016-10-03
Allow vulcan to recognise the new user-agent

david-smith-ft:
Since the user agent was changed in neoutils to describe the exe making the call (eg. "**public-people-api (using neoutils)**"),  the vulcan rewrite rule for public reads hasn't been working correctly on connection.


catalin-ilea:
Let's try to remove HeaderRegexp(`User-Agent`, `neoism|.*neoutils.*`) altogether. Might be wrong but can't see any reason why we need that.

PR #819, opened 2016-10-04
Added JAVA_OPTS to configure memory usuage.

carlosroman:


JuliaFernee:
+1

binary-ingester@.service
```diff
@@ -20,7 +20,9 @@ ExecStart=/bin/sh -c '\
   export KAFKA_PROXY=http://$HOSTNAME:8080; \
   export QUEUE=kafka; \
   export BACKOFF_PERIOD=8000; \
+  export JAVA_OPTS="-Xms296m -Xmx296m -XX:+UseG1GC -server"; \
```
carlosroman: Made these match the binary writer as they'll be processing the same size files.

PR #821, opened 2016-10-04
Update dns ttls

efinlay24:
Bumping elb-dns-registrator to v4.0.1 and tunnel-registrator to v1.0.1 to update DNS TTL values to 30s.


catalin-ilea:
Why do we need 30s on the elb-dns-registrator?

PR #822, opened 2016-10-05
Bumped version to include cypher improvements

david-smith-ft:
https://github.com/Financial-Times/public-people-api/pull/10

### DIFF
--- a/services.yaml
+++ b/services.yaml
@@ -350,7 +350,7 @@ services:
 - name: public-people-api-sidekick@.service
   count: 2
 - name: public-people-api@.service
-  version: v0.1.3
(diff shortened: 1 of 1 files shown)


david-smith-ft:
Load test before: http://ftjen10085-lvpr-uk-p:8181/view/API-load-test/job/people-read-api-load-test/66/console
After: http://ftjen10085-lvpr-uk-p:8181/view/API-load-test/job/people-read-api-load-test/67/console
Highlights are a 49% increase in requests per second and request percentage above 800ms down from 0.048% to 0.0035% (and none above 1200ms).

PR #824, opened 2016-10-06
Added JAVA_OPTS memory constraints to Binary Writer.

carlosroman:


binary-writer@.service
```diff
@@ -23,7 +23,9 @@ ExecStart=/bin/sh -c '\
   export S3_ACCESSKEY=$(/usr/bin/etcdctl get /ft/_credentials/aws/aws_access_key_id); \
   export S3_SECRETKEY=$(/usr/bin/etcdctl get /ft/_credentials/aws/aws_secret_access_key); \
   export BUCKET=$(/usr/bin/etcdctl get /ft/config/binary-writer/bucket); \
+  export JAVA_OPTS="-Xms296m -Xmx296m -XX:+UseG1GC -server"; \
```
khatton-ft: Seems an unusually precise number. I'd have gone up in chunks of 64MB.

carlosroman: I kinda was but I think I added/multiplied incorrectly in my head. I'm now not 100% sure how I ended up at 296m but it works :)

PR #839, opened 2016-10-11
Methode article transformer java opts

JuliaFernee:


khatton-ft:
This is high, but I have often looked at the MAT code and thought it's quite inefficient in the way things are repeatedly converted between DOM objects and Strings, so it's hardly a surprise.

JuliaFernee:
This PR is very old. It was forgotten. Would be a valid one but MAT is deprecated now and is replaced with mapper (MAM)

PR #840, opened 2016-10-12
splunk-forwarder: Switch to Splunk HTTP Event Collector

jussi-ft:
This branch upgrades splunk-forwarder service to version 1.0.0 which includes implementation for Splunk HEC client. 
This release no longer connect to splunk-forwarder cluster but instead sends events directly to Splunk Cloud.

### DIFF
--- a/splunk-forwarder.service
+++ b/splunk-forwarder.service
@@ -20,7 +20,7 @@ ExecStartPre=/bin/bash -c 'docker history coco/coco-logfilter:$DOCKER_LOGFILTER_
 ExecStart=/bin/sh -c '\
   export FORWARD_URL=$(/usr/bin/etcdctl get /ft/config/splunk-forwarder/splunk_hec_url); \
   export TOKEN=$(/usr/bin/etcdctl get /ft/config/splunk-forwarder/splunk_hec_token); \
-  export BATCHSIZE=$(/usr/bin/etcdctl get /ft/config/splunk-forwarder/batchsize) || export BATCHSIZE=10; \
+  BATCHSIZE=$(/usr/bin/etcdctl get /ft/config/splunk-forwarder/batchsize) || BATCHSIZE=10; \
   export ENV=$(/usr/bin/etcdctl get /ft/config/environment_tag); \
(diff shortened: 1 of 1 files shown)


splunk-forwarder.service
```diff
@@ -18,11 +18,12 @@ ExecStartPre=/bin/bash -c 'docker history coco/coco-splunk-http-forwarder:$DOCKE
 ExecStartPre=/bin/bash -c 'docker history coco/coco-logfilter:$DOCKER_LOGFILTER_VERSION >/dev/null 2>&1 \
   || docker pull coco/coco-logfilter:$DOCKER_LOGFILTER_VERSION'
 ExecStart=/bin/sh -c '\
-  export FORWARD_URL=$(/usr/bin/etcdctl get /ft/config/splunk-forwarder/splunk_url); \
+  export FORWARD_URL=$(/usr/bin/etcdctl get /ft/config/splunk-forwarder/splunk_hec_url); \
+  export TOKEN=$(/usr/bin/etcdctl get /ft/config/splunk-forwarder/splunk_hec_token); \
   export ENV=$(/usr/bin/etcdctl get /ft/config/environment_tag); \
```
tamas-molnar: Could you read the batch size from ETCD, so we can more easily play around with it? 
You can have a default value too so it won't fail if the value isn't there, like this: https://github.com/Financial-Times/coco-fleet-deployer/blob/master/deployer.service#L16

tamas-molnar:
Does Splunk Cloud have any rate limiting?

splunk-forwarder.service
jussi-ft: Good idea Tommy. Will implement this and test it in XP.

jussi-ft: Tommy, 
Your suggested change has been implemented here https://github.com/Financial-Times/up-service-files/pull/840/commits/b12b931584c882211314874825aa9fdb4f9566bf. 
Change has been pushed to XP delivery cluster and tested tuning the batchsize value in etcd. Working as expected.

PR #852, opened 2016-10-18
Add concept suggestion api to cluster

scott-ace-newton:


concept-suggestion-api@.service
```diff
@@ -0,0 +1,30 @@
...
+# Change killmode from "control-group" to "none" to let Docker remove work correctly.
+KillMode=none
+ExecStartPre=-/bin/bash -c '/usr/bin/docker kill "$(docker ps -q --filter=name=%p-%i_)" > /dev/null 2>&1'
+ExecStartPre=-/bin/bash -c '/usr/bin/docker rm "$(docker ps -q --filter=name=%p-%i_)" > /dev/null 2>&1'
+ExecStartPre=/bin/bash -c '/usr/bin/docker history up-registry.ft.com/coco/concept-suggestion-api:$DOCKER_APP_VERSION > /dev/null 2>&1 \
+  || /usr/bin/docker pull up-registry.ft.com/coco/concept-suggestion-api:$DOCKER_APP_VERSION'
+
+ExecStart=/bin/sh -c '\
+  export APP_PORT=8080; \
+  export CES_ENDPOINT="http://ftapp06179-lvpr-uk-t:8080/worker/extract" \
```
tosan88: This host looks like it's a test machine and not public accessible.... Is this correct?

concept-suggestion-api@.service
```diff
@@ -0,0 +1,32 @@
...
+# Change killmode from "control-group" to "none" to let Docker remove work correctly.
+KillMode=none
+ExecStartPre=-/bin/bash -c '/usr/bin/docker kill "$(docker ps -q --filter=name=%p-%i_)" > /dev/null 2>&1'
+ExecStartPre=-/bin/bash -c '/usr/bin/docker rm "$(docker ps -q --filter=name=%p-%i_)" > /dev/null 2>&1'
+ExecStartPre=/bin/bash -c '/usr/bin/docker history up-registry.ft.com/coco/concept-suggestion-api:$DOCKER_APP_VERSION > /dev/null 2>&1 \
+  || /usr/bin/docker pull up-registry.ft.com/coco/concept-suggestion-api:$DOCKER_APP_VERSION'
+
+ExecStart=/bin/sh -c '\
+  export APP_PORT=8080; \
+  export CES_ENDPOINT="$(/usr/bin/etcdctl get /ft/_credentials/ces/endpoint)"; \
```
tosan88: Can you please place it in `/ft/config/` instead of `/ft/_credentials`?

david-smith-ft: I put it in `/ft/_credentials` because the endpoint contains basic auth in the URL.  I could split it out into individual entries but they're never used separately so I didn't see a reason to.

PR #867, opened 2016-10-27
Set Filter Novelties to true

scott-ace-newton:
This value is no longer hard coded in the dockerfile.


scott-ace-newton:
http://git.svc.ft.com:8080/projects/CP/repos/content-annotator/commits/bc8dbc36a606425e8ece33a7cdff3fca60773ede

PR #872, opened 2016-11-04
Add rec reads content annotator

scott-ace-newton:
Adding Recommended Reads content annotator to coco clusters as we migrate suggestions flow from non-coco. This service will read ConceptSuggestions topic and send annotations to rr-suggestions-indexer in ftp2 via varnish and elb.

### DIFF
--- a/rec-reads-content-annotator-sidekick@.service
+++ b/rec-reads-content-annotator-sidekick@.service
@@ -1,0 +1,25 @@
+[Unit]
+Description=Recommended Reads Content Annotator Service Sidekick
+BindsTo=rec-reads-content-annotator@%i.service
+After=rec-reads-content-annotator@%i.service
+
+[Service]
+RemainAfterExit=yes
+ExecStart=/bin/sh -c "\
+  export SERVICE=$(echo %p | sed 's/-sidekick//g'); \
+  etcdctl set   /ft/services/$SERVICE/healthcheck     true; \
+  etcdctl mkdir /ft/services/$SERVICE/servers; \
+  etcdctl set /ft/healthcheck/$SERVICE-%i/path /__health; \
+  while [ -z $PORT ]; do \
+    sleep 5; \
+    CONTAINER_NAME=$(docker ps -q --filter=name=\"$SERVICE\"-%i_); \
+    PORT=`echo $(/usr/bin/docker port $CONTAINER_NAME 8080 | cut -d':' -f2)`; \
+  done; \
+  etcdctl set /ft/services/$SERVICE/servers/%i http://%H:$PORT;"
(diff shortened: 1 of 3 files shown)


rec-reads-content-annotator-sidekick@.service
```diff
@@ -0,0 +1,25 @@
+[Unit]
+Description=Recommended Reads Content Annotator Service Sidekick
+BindsTo=rec-reads-content-annotator-sidekick@%i.service
+After=rec-reads-content-annotator-sidekick@%i.service
```
tosan88: `BindsTo` & `After` should have the service instead of the sidekick itself

rec-reads-content-annotator@.service
```diff
@@ -0,0 +1,47 @@
...
+  export APP_PORT=8080; \
+  export GROUP_ID=$(/usr/bin/etcdctl get /ft/config/environment_tag)"-rec-reads-content-annotator"; \
+  export ZOOKEEPER_HOST=$(/usr/bin/etcdctl get /ft/config/zookeeper/ip); \
+  export ZOOKEEPER_PORT=$(/usr/bin/etcdctl get /ft/config/zookeeper/port); \
+  export KAFKA_HOST=$(/usr/bin/etcdctl get /ft/config/kafka/ip); \
+  export KAFKA_PORT=$(/usr/bin/etcdctl get /ft/config/kafka/port); \
+  export JAVA_OPTS="-Xms256m -Xmx256m -XX:+UseG1GC -server"; \
+  export WRITER_HOST=$(/usr/bin/etcdctl get /ft/config/upp-gateway/host); \
+  export WRITER_PORT=$(/usr/bin/etcdctl get /ft/config/upp-gateway/port); \
+  export WRITER_CREDENTIALS=$(usr/bin/etcdctl get /ft/config/upp-gateway/credentials); \
```
tosan88: I would expect credentials to be under `/ft/_credentials/...`

tosan88: Also, assuming these will be configured in provisioner as well :)

rec-reads-content-annotator-sidekick@.service
scott-ace-newton: Fixed

rec-reads-content-annotator@.service
scott-ace-newton: Fixed and I will add this to the provisioner in a bit

scott-ace-newton:
The version differences between our other content annotators and this one can be found here by looking at the last 7 commits http://git.svc.ft.com:8080/projects/CP/repos/content-annotator/commits

PR #876, opened 2016-11-07
Neo rw batchsize config

JuliaFernee:
The service file sets key in etcd only once to a default 600 if the key doesn't exist.  Later we can change this value in etcd as needed, restart the service the new value will be picked up. Note that etcd holds keys per each service instance, this means that the values /ft/services/organisations-rw-neo4j-red-instance-1/param/batchsize and /ft/services/organisations-rw-neo4j-blue-instance-1/param/batchsize can be different and so may both  need to be updated.


JuliaFernee:
Removed changes to the wordpress-article-transformer.

people-rw-neo4j-blue@.service
```diff
@@ -20,7 +22,7 @@ ExecStart=/bin/sh -c '\
   --memory="256m" \
   --env "NEO_URL=http://$HOSTNAME:8080/__neo4j-blue/db/data" \
   --env "APP_PORT=$APP_PORT" \
-  --env "BATCH_SIZE=600" \
+  --env "BATCH_SIZE=$BATCH_SIZE" \
```
tosan88: Getting BATCH_SIZE variable is missing? Also in `people-rw-neo4j-red@.service`, `roles-rw-neo4j-blue@.service`, `roles-rw-neo4j-red@.service`, `sections-rw-neo4j-blue@.service`, `sections-rw-neo4j-red@.service`, `special-reports-rw-neo4j-blue@.service`,`special-reports-rw-neo4j-red@.service`

PR #878, opened 2016-11-10
Syndicated content

kovacshuni:
* https://github.com/Financial-Times/api-policy-component/pull/1
* http://git.svc.ft.com/projects/CP/repos/methode-article-transformer/pull-requests/112/overview
* http://git.svc.ft.com/projects/CP/repos/methode-article-transformer/pull-requests/111/overview
* http://git.svc.ft.com/projects/CP/repos/content-public-read/pull-requests/26/overview

the URLs for methode-api are changed to be pointing to TEST but be careful, in prod we shouldn't change that. I will take care of that- huni


kovacshuni:
I need to change the input env vars SOURCE_APP_URI and SOURCE_APP_HEALTH_URI to get their values from etcd, as they have different values in different environments. This way of static reference in service files doesn't work. It will revert the configs in lower environments on the first time master is merged back to xp or so. I decided we go ahead to prod now, as it's not affecting prod, and the functionality is needed, it's been too long we've waited for this. I'll fix after.

PR #879, opened 2016-11-11
Add rec reads ingester

scott-ace-newton:
Adding Recommended Reads content ingester to coco clusters as we migrate suggestions flow from non-coco. This service will read CmsPublicationEvents topic and send content to rr-content-indexer in ftp2 via varnish and an elb.


scott-ace-newton:
Differences between current ingester and this can be found http://git.svc.ft.com:8080/projects/CP/repos/ingester/commits. Mainly needed to add basic auth for varnish and enabling https

PR #900, opened 2016-11-28
Set concept suggestion uri correctly

scott-ace-newton:


scott-ace-newton:
http://git.svc.ft.com:8080/projects/CP/repos/concept-suggestion-api/commits/ceed4edf2940f3ffc9e2ded2aa20611d9f2a1614

PR #905, opened 2016-11-30
changed the bridges so that they read from both publishing US and UK

yolandeleungFT:
**for preprod only**
still need to do concept suggestions

### DIFF
--- a/services.yaml
+++ b/services.yaml
@@ -47,9 +47,14 @@ services:
 - name: burrow@.service
   version: v1.0.1
   count: 1
-- name: cms-kafka-bridge-pub-pre-prod-sidekick@.service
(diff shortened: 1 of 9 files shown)


services.yaml
```diff
@@ -47,9 +47,14 @@ services:
 - name: burrow@.service
   version: v1.0.1
   count: 1
-- name: cms-kafka-bridge-pub-pre-prod-sidekick@.service
+- name: cms-kafka-bridge-pub-pre-prod-uk-sidekick@.service
   count: 2
-- name: cms-kafka-bridge-pub-pre-prod@.service
+- name: cms-kafka-bridge-pub-pre-prod-uk@.service
+  version: v13.0.0
```
izzyblues: I didn't understand why you did not put version v14.0.0

PR #913, opened 2016-12-06
Feature/content notification push

izzyblues:


notifications-push@.service
```diff
@@ -27,6 +27,8 @@ ExecStart=/bin/sh -c '\
       --env "QUEUE_HOST=kafka" \
       --env "API_BASE_URL=$API_BASE_URL" \
       --env "CONSUMER_BACKOFF=2" \
+      --env "NOTIFICATIONS_RESOURCE=content" \
+      --env "WHITELIST=^http://.*-transformer-(pr|iw)-uk-.*\.svc\.ft\.com(:\d{2,5})?/(content)/[\w-]+.*$" \
```
Nesze: If you could add the mapper here also, it would be great.

izzyblues: What's the mapper URL look like?

Nesze: Just adding an OR to transformer should work

> ...-(transformer|mapper)-...

izzyblues: 👍

PR #922, opened 2016-12-13
Add concept load flow

BerniVarga:
- Adding service files for concept-rw-elasticsearch
- Upgrading concept-ingester to a version, which support writes into elasticsearch
- Modifying the concept-ingester-neo4j-blue to write into elasticsearch
- Applying format.go to services.yaml (it needed already:)

Tested in XP. :)


tosan88:
Is there any reason why you didn't change `concept-ingester-neo4j-red` as well?

BerniVarga:
Yes, we didn't want to double the load on elasticsearch. 
Also, replacing the 2 neos with a neo cluster is in plan already, so hopefully we won't stick with this workaround for too long.

kovacshuni:
+1

PR #925, opened 2016-12-15
Fix/notif push uuid parsing

Nesze:
* https://github.com/Financial-Times/notifications-push/pull/35
* fix whitelist regexp


Nesze:
Tested in XP.

Nesze:
Tested in pre-prod.

PR #936, opened 2016-12-15
remove cms bridges, which were moved to pub clusters

Nesze:
Edit: It's debatable whether this makes pre-prod to diverge even more from prod. Let me know what you think.


Nesze:
Tried it, then find out that it's not worth it, so we backed off.

PR #942, opened 2017-01-04
notifications: fix units' stop cmd

Nesze:
Stopping `notifications-push` has not been working since `list-notifications-push` was deployed. It turns out that container filtering in case of overlapping unit names is not restrictive enough.

The same could happen with `notifications-rw` if `list-notifications-rw` runs on the same machine.

This could happen with any new overlapping unit name.

### DIFF
--- a/notifications-push@.service
+++ b/notifications-push@.service
@@ -9,8 +9,8 @@ Environment="DOCKER_APP_VERSION=latest"
 TimeoutStartSec=0
 # Change killmode from "control-group" to "none" to let Docker remove work correctly.
 KillMode=none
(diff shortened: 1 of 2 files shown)


Nesze:
Tested in pre-prod.

tosan88:
Just for the record: there is a simpler solution documented in the template: https://github.com/Financial-Times/up-service-files/blob/master/template/template%40.service#L40

Nesze:
That's way more elegant! It should be mentioned in the `docker ps` doc, though...

I will update this in the upcoming days.

PR #948, opened 2017-01-05
Fixing filtering for docker names collision around notifications

tosan88:
Follow up on https://github.com/Financial-Times/up-service-files/pull/942 AND fixes for services which were left out.

Tested in xp


tosan88:
Works in pre-prod as expected

PR #949, opened 2017-01-06
Fixed issue regarding notifications-ingester stop...

izzyblues:
…when list-notifications-ingester is on the same machine


izzyblues:
Already fixed in https://github.com/Financial-Times/up-service-files/pull/948

PR #953, opened 2017-01-10
Fix delete organisation rw

scott-ace-newton:
Affected code changes here:
https://github.com/Financial-Times/organisations-rw-neo4j/pull/21
and here:
https://github.com/Financial-Times/organisations-rw-neo4j/pull/22


scott-ace-newton:
Works in semantic

PR #957, opened 2017-01-12
Increase kafka proxy message size

danielcopaciu:
This change is needed for making kafka-proxy able to handle image binary messages.


kafka-rest-proxy@.service
```diff
@@ -22,6 +22,7 @@ ExecStart=/bin/bash -c "\
   -e RP_FETCH_MESSAGE_MAX_BYTES=16777216 \
   -e RP_CONSUMER_REQUEST_MAX_BYTES=17825792 \
   -e RP_ZOOKEEPER_CONNECT=%H:2181 \
+  -e RP_MAX_REQUEST_SIZE=16777216 \
```
Nesze: Shouldn't this be also then `17825792` ?

PR #978, opened 2017-01-20
Bugfix/ingester oom logging

izzyblues:


khatton-ft:
Tested in DynPub.

khatton-ft:
@bogdanguranda All deployments to the cluster use the config-docker.yaml file and as you say, the service file sets up message forwarding. The for-content-store.yaml file is a convenience for local development and I realised it no longer matched the intended behaviour for the content ingester (and in fact it was masking a problem), so I just updated it as part of the work.

PR #990, opened 2017-01-27
Fix/mlm map endpoint

Nesze:
See:
http://git.svc.ft.com/projects/CP/repos/methode-list-mapper/pull-requests/6/overview

Tested in XP. 

PS: Sorry about the branch name, this is rather a feature.


horia-giurgiu-ft:
Tested in XP and PRE-PROD for:
A. Happy flow
B. Bad flow
1. if ObjectMetadata/Lists/StreamID is not empty then ObjectMetadata/Lists/ListType should be not empty
2. you can't have in document store two lists with the same streamID and listType
3. ObjectMetadata/FTcom/DIFTcomWebType should be only digitalList

PR #985, opened 2017-01-25
Updated the version of MAM with fix to not throw errors for Content Placeholders

ppragnya:

PR #993, opened 2017-01-30
Fix/mam map endpoint

Nesze:
See https://github.com/Financial-Times/methode-article-mapper/pull/10

Note:

* More on v9 check https://github.com/Financial-Times/up-service-files/pull/985

Again, sry for the branch name, it's more like a feature, rather than a fix.


ancabusuioc:
Tested in xp.

PR #995, opened 2017-02-02
Updated to use lifcycle version

david-smith-ft:


brightcove-annotations-rw-blue@.service
```diff
@@ -20,6 +20,7 @@ ExecStart=/bin/sh -c '\
   /usr/bin/docker run --rm --name %p-%i_$(uuidgen) -p $APP_PORT \
   --memory="256m" \
   --env "PLATFORM_VERSION=brightcove" \
+  --env "LIFECYCLE=brightcove-annotations" \
```
iulianp26: In app.go is expected an env variable named: ANNOTATIONS_LIFECYCLE

david-smith-ft:
@iulianp26 You're completely correct, my mistake.  I've updated the service files accordingly.

david-smith-ft:
Red and blue have gone

PR #997, opened 2017-02-06
Methode article internal components mapper

Nesze:
See: 

https://github.com/Financial-Times/methode-article-internal-components-mapper

Tested in `rj`.


Nesze:
Tested in `rj`.

Nesze:
Added the JVM memory options & updated the sidekick conform to the template.

Nesze:
Deployed to pre-prod.

Nesze:
Tested in pre-prod.

PR #999, opened 2017-02-06
Disabled the timeout logic for SystemD killing Neo4J

nwrigh:
Disabled the timeout logic for killing processes to avoid corruption of the datastore due to systemd killing the Neo instance. The systemd default is 90s before it issues a SIGKILL and the datastore may have grown just enough that it takes longer than 90s. Setting to infinity allows the service to shutdown in its own time and so a nasty shutdown and subsequent corruption should be avoided

http://unix.stackexchange.com/questions/227017/how-to-change-systemd-service-timeout-value


tosan88:
I thought this [would kill the container after 3 seconds](https://docs.docker.com/engine/reference/commandline/stop/#/options): ```ExecStop=-/bin/bash -c 'docker stop -t 3 "$(docker ps -q --filter=name=^/%p-%i_)"'```

PR #1001, opened 2017-02-07
Internal Content API service

radutudor:
https://github.com/Financial-Times/internal-content-api/pull/1
https://github.com/Financial-Times/internal-content-api/pull/3
https://github.com/Financial-Times/internal-content-api/releases/tag/v0.0.5


Nesze:
Tested in pre-prod.

Note: the endpoint has not been load tested yet. We put this out in prod, though. 
Next will get the api keys only after it is load tested.

PR #1004, opened 2017-02-08
Varnish version bump to 0.1.7

radutudor:
New Varnish version which has request limit set for internalcontent-preview
https://github.com/Financial-Times/delivery-varnish/pull/9


Nesze:
Tested in pre-prod.

PR #1017, opened 2017-02-15
Internal components mapper, changed bgColor label to backgroundColour 

bogdanguranda:
As decided with Sarah, the internal component topper will expose the label backgroundColour label instead of bgColor.

https://github.com/Financial-Times/methode-article-internal-components-mapper/pull/6


Nesze:
I think we should also try to republish all articles with a topper, so we are consistent with the `backgroundColour` naming.

PR #1019, opened 2017-02-15
Wordpress article mapper map endpoint

danielcopaciu:
https://github.com/Financial-Times/wordpress-article-mapper/pull/7


ancabusuioc:
Verfied in xp and pre-prod.

PR #1030, opened 2017-02-21
Parameterised notification since date in load replicator

scott-ace-newton:
https://github.com/Financial-Times/load-replicator/pull/5


scott-ace-newton:
Tested locally, pointing to semantic

PR #1032, opened 2017-02-22
Bump load replicator to v1.0.2

scott-ace-newton:
https://github.com/Financial-Times/load-replicator/pull/5
https://github.com/Financial-Times/load-replicator/pull/6


scott-ace-newton:
supposed to be to pre-prod not master

PR #1036, opened 2017-02-23
Add sequential deployment for MAM.

khatton-ft:
Due to content-preview, MAM is effectively in the read category, and
therefore it should be deployed sequentially to ensure read remains GTG
throughout.


izzyblues:
@khatton-ft Is this an issue that we have only in dynpub? Otherwise I do not understand why you created a PR to merge in dynpub.

khatton-ft:
@izzyblues My understanding was that there should be a PR for the lower environment, for anything that is not just a version number change.

Nesze:
Could you pls set this to the `methode-article-internal-components-preview` service, too?

That's now also part of the (internal) preview flow.

Nesze:
Cheers

PR #1041, opened 2017-02-24
SwitchElasticSearchConceptIngester

scott-ace-newton:
Elastic search should be fed from neo cluster concept ingester not from soon to be deprecated blue concept ingester


scott-ace-newton:
tested in pre-prod

PR #1044, opened 2017-02-27
Map instead content transform

kovacshuni:
https://github.com/Financial-Times/content-preview/pull/19


Nesze:
Please also update the internal-content-preview unit.

PR #1055, opened 2017-03-02
Removed all red/blue rws, annotators and ingesters service files

scott-ace-newton:
Removed all red/blue rws, annotators and ingesters service files and records from services.yaml. Also bumped version of content-ingester-neo4j to v65 as it must have been missed from a previous pr.


scott-ace-newton:
Tested in semantic

scott-ace-newton:
@tosan88 Not really. This pr should only be merged after the neo reads are pointed at the new Neo4j clusters

PR #1064, opened 2017-03-06
Added env vars to make configurable healthcheck values

bogdanguranda:
* 2 of this services have the same codebase and the other 2 have another same codebase which resulted in wrong panic guide urls and information in the healthcheck for the newer services.

* Added env vars will not be processed until the changes on the services will also be released, but this PR will not break the services:
https://github.com/Financial-Times/internal-content-api/pull/8
https://github.com/Financial-Times/content-preview/pull/21


bogdanguranda:
@Nesze thx for noticing. There are actually no intermediate versions, I've just bumped the current prod version with 3 version number because of a misunderstanding, but this is just one version away from prod.

PR #1066, opened 2017-03-07
bump enriched-content-read-api to new version that support story and content packages

radutudor:
http://git.svc.ft.com/projects/CP/repos/enriched-content-read-api/pull-requests/28/overview


kovacshuni:
it's in pre-prod, in prod is still v70

radutudor:
You're right! I actually checked during the standup but forgot to mention it ...

PR #1068, opened 2017-03-08
Added content-collection-rw-neo4j and content-collection-ingester services

ioanaciont:
…vices


kovacshuni:
ok, so since everyone started using release candidate tags, according to semver.org, you can only use dash and plus symbol once in a version. so no teamname-rc-1. I put this here as an FYI for the future.

tosan88:
@kovacshuni Thanks, I think we didn't consider this when we agreed. Time to think about it?

kovacshuni:
@tosan88 sure

kovacshuni:
Let's fall into convention with the versions in all the platform, as the majority of versions is with a v prefix, could you rebuild this to be v1.0.0? And if you can the neo one as well, add .0.0? If it's a jenkins thing and it's too much hussle, you can leave that.

tosan88:
@kovacshuni I would rather not do that, as all the other teams are releasing without the v prefix which complies with semver. I encourage to discuss this more widely though, maybe in UPP tech forum?

kovacshuni:
ok, go with this as is, but let's propose versions at tech forum.

PR #1082, opened 2017-03-17
added internalcontent-read category

Nesze:
(a corresponding dashing tile will follow soon)


Nesze:
Tested in `rj`.

Nesze:
I think we only want services of platinum endpoints to be included here.

PR #1095, opened 2017-03-24
Fix/type

Nesze:
1. [ContentPackage support for notifications](http://git.svc.ft.com/projects/CP/repos/notifications-rw/pull-requests/71/overview)
2. Setting the specific type in backend and prepend the type hierarchy on the read side: [MAM](https://github.com/Financial-Times/methode-article-mapper/pull/24) / [CPR](http://git.svc.ft.com/projects/CP/repos/content-public-read/pull-requests/38/overview)

Tested in `rj`.

Note: ContentPackages will need to be republished right after the release, as they already have the full type saved in the backend.


Nesze:
Tested in `pre-prod`.

PR #1098, opened 2017-03-27
Using v1.1 of go-fthealth lib + vendoring

bogdanguranda:
**Using new v1.1 go-fthealth library.
Added vendoring (govendor).**

*Updated services:*
- internal-content-api
- internal-content-preview-api
- content-preview
- internal-components-preview
- content-collection-rw-neo4j

*Codebase PRs:*
https://github.com/Financial-Times/internal-content-api/pull/15
https://github.com/Financial-Times/content-preview/pull/22
https://github.com/Financial-Times/content-collection-rw-neo4j/pull/4


internal-content-api@.service
```diff
@@ -27,7 +27,8 @@ ExecStart=/bin/sh -c '\
     ENV_API_HOST=$(etcdctl get /ft/config/api_host); \
     /usr/bin/docker run --rm --name %p-%i_$(uuidgen) -p 8080 -p 8081 \
     --memory="256m" \
-    --env "APP_NAME=internal-content-api" \
+    --env "APP_SYSTEM_CODE=up-ica" \
```
radutudor: Why is this `up-ica` when all the rest of the SYSTEM CODES look like: `internal-content-preview-api`, `internal-components-preview` etc etc

Shouldn't this be `internal-content-api` ?

content-collection-rw-neo4j@.service
```diff
@@ -22,12 +22,14 @@ ExecStart=/bin/sh -c '\
   export NEO_URL=$(/usr/bin/etcdctl get /ft/config/neo4j/read_write_url); \
   /usr/bin/docker run --rm --name %p-%i_$(uuidgen) -p $APP_PORT \
   --memory="256m" \
-  -e "NEO_URL=$NEO_URL" \
-  -e "APP_PORT=$APP_PORT" \
-  -e "GRAPHITE_ADDRESS=graphite.ft.com:2003" \
-  -e "GRAPHITE_PREFIX=coco.services.$ENV.content-collection-rw-neo4j.%i" \
-  -e "LOG_METRICS=false" \
-  -e "CACHE_DURATION=30s" \
+  --env "APP_SYSTEM_CODE=upp-content-collection-rw-neo4j" \
```
radutudor: why does this have `upp` in the SYSTEM CODE when most other app do not?

internal-content-api@.service
kovacshuni: https://dewey.ft.com/ system code column, not conventional, i know.

content-collection-rw-neo4j@.service
bogdanguranda: System Codes are not consistent with a standard, there are old & new naming conventions (I guess this is new).
That's how it is defined in https://dewey.ft.com/
I'm can't change that. I know there were discussions on the convention, but I don't know who is in charge of this decisions and if it is worth changing now.
Same for the other question.

PR #1100, opened 2017-03-29
Save content type to cph

yolandeleungFT:


kovacshuni:
great, thank you for this! - team rj :)

PR #1181, opened 2017-05-30
Tie the mongo backup timer to the mongo backup service

tamas-molnar:
- when the cluster is restarted - or just the persistent machines - the `mongo-backup@1.timer` is in an unhealthy state and needs a kick to recover
- I found nothing in the logs to suggest an error, it just wasn't started for some reason, more details on UPP2-192
- suspecting a dependency issue, I tied the timer to the machines the backup service is on, rather than the mongos, so it starts up AFTER the backup services, and from the testing in XP so far, it worked, I could restart all persistent machines with the `mongo-backup@1.timer` starting up cleanly
- to release this, push the changes to GH, delete all timers from the cluster and kick the deployer

### DIFF
--- a/mongo-backup@.timer
+++ b/mongo-backup@.timer
@@ -8,4 +8,4 @@ OnCalendar=11:00
 WantedBy=timers.target
 
 [X-Fleet]
-MachineOf=mongodb@%i.service
+MachineOf=mongo-backup@%i.service


tamas-molnar:
- the fix which worked in team envs didn't work the same way in pre-prod, so this may still appear
- we were discussing removing backups from lower envs so this may go away with that, or when we get a proper clustering and backup/restore solution for mongo
- this change was deployed up to pre-prod, I'll leave it there for now

PR #1183, opened 2017-06-01
Change bucket region etcd key for aggreggate-concept-transformer

david-smith-ft:
Ensure it's using a key under the /ft/config/aggregate-concept-transformer config location.

`/ft/config/aggregate-concept-transformer/bucket_region` will need to be populated before merging.


david-smith-ft:
Included in the full brands roll-out.

PR #1190, opened 2017-06-07
content-rw-elasticsearch with uuid lib

radutudor:


bogdanguranda:
This was handed over to another team and integrated in their changes: https://github.com/Financial-Times/content-rw-elasticsearch/pull/6

PR #1236, opened 2017-06-29
Disable Splunk Forwarder

efinlay24:
Due to Splunk Cloud license limits, this disables the Splunk Forwarder on a temporary basis for the `pre-prod` environments until we can get more permanent code changes out.


efinlay24:
Nope - `splunk-forwarder` simply tails the log output from when it's started. When we start it back up again, it'll simply start forwarding again from that point in time. :)

PR #1250, opened 2017-07-07
Debug/concept rw es profiling

izzyblues:
Added profiling to concept-rw-elasticsearch.
See: https://github.com/Financial-Times/concept-rw-elasticsearch/pull/22


kovacshuni:
can be closed, right? 1.1.0 is in production.

khatton-ft:
Now obsolete.

PR #1269, opened 2017-07-17
Feature/concept search reindexer

khatton-ft:


khatton-ft:
NB This can be reviewed now, but shouldn't be merged until the version of the index mapping has reached the current version.

PR #1295, opened 2017-07-31
cleaning up pac poc apps

galiri:


services.yaml
```diff
@@ -173,11 +167,6 @@ services:
   version: v95
   count: 2
   sequentialDeployment: true
-- name: draft-content-rw-aurora-sidekick@.service
```
khatton-ft: The service / sidekick files for this service can also be removed.

services.yaml
```diff
@@ -156,12 +156,6 @@ services:
 - name: content-rw-neo4j@.service
   version: 1.0.6
   count: 2
-- name: content-rw-s3-sidekick@.service
```
khatton-ft: The annotations-rw-rdb and annotations-rw-s3 services can also be removed from this file.

services.yaml
galiri: done

services.yaml
galiri: I cannae see them to remove??

PR #1298, opened 2017-08-01
Bumped version of api that handles new concordance model

scott-ace-newton:
https://github.com/Financial-Times/public-content-by-concept-api/pull/10

Tested in semantic


scott-ace-newton:
Weve reverted back to having these all released in 1 pr

PR #1311, opened 2017-08-09
Concepts-rw-neo4j - Adding an unique constraint

scott-ace-newton:
https://github.com/Financial-Times/concepts-rw-neo4j/pull/8

Tested in semantic


scott-ace-newton:
👍

PR #1315, opened 2017-08-09
api-policy-component:64.4.2-rj-norecommended-rc1 no recommended

kovacshuni:
https://github.com/Financial-Times/api-policy-component/pull/24

Tested recommended tags appearance/non-appearance in rj's endpoints and database collections.
More to test in pre-prod.


kovacshuni:
We should test if the `suppressMarkup` filters are still correct, it was a little more worrying issue during merging.

PR #1319, opened 2017-08-16
Fix/cpr no uuid params

scott-ace-newton:
Tested in Semantic

https://github.com/Financial-Times/document-store-api/pull/2

https://github.com/Financial-Times/content-public-read/pull/2

### DIFF
--- a/services.yaml
+++ b/services.yaml
@@ -146,13 +146,13 @@ services:
 - name: content-public-read-preview-sidekick@.service
   count: 2
 - name: content-public-read-preview@.service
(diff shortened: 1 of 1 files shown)


khatton-ft:
The Jenkins log for CPR build 86 suggests there was a failure. Have you checked that? So long as the Docker image was built successfully, it's OK.

scott-ace-newton:
Yeah I looked into that and its to do with the execute shell command at the bottom which I think was to remove the image from the jenkins box but that didnt make much sense to me? I was going to ask someone but forgot. Either way the image built fine and is in the registry

PR #1325, opened 2017-08-23
Use GLB address for load-replicator

efinlay24:
Currently load-replicator sends traffic to the cluster address (eg, `pre-prod-uk-up.ft.com`), which doesn't simulate failovers correctly.

This change sends all load-replicator traffic via the GLB `pre-prod-up.ft.com` address.


khatton-ft:
This change would mean we can _only_ use the load replicator in pre-prod ... have we considered that?

efinlay24:
Yeah, unfortunately it does mean that:
- the service file will diverge between branches
- people will need to remember to update the service file if necessary when copying to a new environment

...but I think that the other options (using a different etcd key, or parse out `-uk` and `-us` from cluster names) add more complexity and different problems.

I'll send out a note to let people know about these changes in case they need to configure the load replicator in different environments.

PR #1328, opened 2017-08-25
Allow push clients to specify type, default to 'only articles'

ioanaciont:
These changes will affect **_new-notifications-push_** service. Clients can filter the content adding a request param for content type (e.g. https://api.ft.com/content/notifications-push?type=article)

- If the type is not specified, the filter will be set to `type=Article`
- Clients can specify `type=All` that will return all supported notifications, including CPHs.
- `DELETE` notifications will be always sent to the clients, because their payload is empty (in Kafka) and we cannot extract the type
- If the client specifies an invalid type, a client error is returned

- App changes: https://github.com/Financial-Times/notifications-push/pull/50 && https://github.com/Financial-Times/notifications-push/pull/51


ioanaciont:
[Avoid notifying push clients about content they can't use](https://trello.com/c/s1Fg6LxJ/38-avoid-notifying-push-clients-about-content-they-cant-use)

PR #1335, opened 2017-09-01
Update roles rw v0.1.7

scott-ace-newton:
https://github.com/Financial-Times/roles-rw-neo4j/pull/10


scott-ace-newton:
Tested in semantic

PR #1345, opened 2017-09-06
Restrict current notifications-push access via Fastly

tosan88:
Please see https://github.com/Financial-Times/delivery-varnish/pull/17


tosan88:
[Forbid calls to current-NP via api.ft.com](https://trello.com/c/Nd3FWS43/40-forbid-calls-to-current-np-via-apiftcom)

PR #1349, opened 2017-09-11
Remove retired brands pipeline

scott-ace-newton:


scott-ace-newton:
Writing of brands to Neo4j is now handled by concepts-rw-neo4j, verified in semantic cluster

PR #1351, opened 2017-09-11
Return all parents children brands api

scott-ace-newton:


scott-ace-newton:
https://github.com/Financial-Times/public-brands-api/pull/15

PR #1369, opened 2017-09-25
Added helm integration and sirupsen changes to membership rw neo4j

nwrigh:
https://github.com/Financial-Times/memberships-rw-neo4j/pull/12
https://github.com/Financial-Times/memberships-rw-neo4j/pull/11

### DIFF
--- a/services.yaml
+++ b/services.yaml
@@ -263,7 +263,7 @@ services:
 - name: memberships-rw-neo4j-sidekick@.service
   count: 2
 - name: memberships-rw-neo4j@.service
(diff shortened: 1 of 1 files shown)


tamas-molnar:
Awe we still using versions with `v`?

nwrigh:
@tamas-molnar We were supposed to have migrated away from this but I realised we hadn't fixed that until after I created a release I decided to weigh up fixing it and the fact that this service will no longer exist by the end of the year or early Q1 and it certainly won't be actively worked on

PR #1378, opened 2017-09-27
Add k8s bridge versions - without host header routing.

BerniVarga:
Will need to be updated with the final tagged version (after merging [this PR](https://github.com/Financial-Times/coco-kafka-bridge/pull/44)).


kovacshuni:
Okay! ✅

PR #1385, opened 2017-09-28
Fix creation of build-info values.

khatton-ft:


services.yaml
```diff
@@ -493,7 +493,7 @@ services:
 - name: wordpress-article-mapper-sidekick@.service
   count: 2
 - name: wordpress-article-mapper@.service
-  version: 1.1.0
+  version: 1.1.1-k8s-build-fix-rc1
```
izzyblues: Do you want to deploy a release candidate version in pre-prod or prod?

PR #1386, opened 2017-09-28
Build from released version

scott-ace-newton:
https://github.com/Financial-Times/aggregate-concept-transformer/pull/14


scott-ace-newton:
Tested in semantic

PR #1389, opened 2017-09-29
Public things api1.5.1

scott-ace-newton:


scott-ace-newton:
https://github.com/Financial-Times/public-things-api/pull/10

PR #1391, opened 2017-09-29
Deploy released version

scott-ace-newton:


scott-ace-newton:
https://github.com/Financial-Times/concepts-rw-neo4j/pull/17

PR #1394, opened 2017-09-29
Bump version to released version

scott-ace-newton:


scott-ace-newton:
https://github.com/Financial-Times/generic-rw-s3/pull/13

PR #1401, opened 2017-10-02
Fix/content rw neo live blogs

peteclark-ft:


peteclark-ft:
If Prod K8s wasn't the same as Pre-Prod, I'd probably wait before pushing there 😳

PR #1423, opened 2017-10-05
New concordance model public people api

scott-ace-newton:


scott-ace-newton:
https://github.com/Financial-Times/public-people-api/pull/16

PR #1452, opened 2017-10-26
Increased service count for CPR

tosan88:
2 -> 3


tosan88:
Tested in `video` cluster

PR #1470, opened 2017-11-13
Concept exports etcd keys

ioanaciont:
Removing `/ft/config/upp-exports-rw-s3/aws_region` and reading the same value from `/ft/config/bucket_region`


tosan88:
Included in https://github.com/Financial-Times/up-service-files/pull/1498

PR #1511, opened 2018-01-04
Run Load replicator only during weekdays and work hours

JermilaPaulDhas:


khatton-ft:
What's the underlying reason for doing this? Don't we want to soak test things with load overnight?

PR #1519, opened 2018-01-09
Load replicator crontime

JermilaPaulDhas:


load-replicator@.service
```diff
@@ -18,7 +18,7 @@ ExecStart=/bin/sh -c '\
   AUTHORIZATION=$(/usr/bin/etcdctl get /ft/config/load-replicator/authorization); \
   NOTIFICATIONS_DATE=$(/usr/bin/etcdctl get /ft/config/load-replicator/notifications); \
   ENVIRONMENT=pre-prod; \
-  DURATION=$(/usr/bin/etcdctl get /ft/config/load-replicator/duration) || DURATION=0; \
+  DURATION=11; \
```
MihaiSorinMoisa: I would use the etcd configuration to set the duration.

PR #1566, opened 2018-01-12
:rocket: RC of public-content-by-concept for helm day III

izzyblues:


scott-ace-newton:
Tested in semantic and k8s-dev

PR #1582, opened 2018-01-18
Replace delete notifications with update notifications

MihaiSorinMoisa:
Send an update notification instead of a delete notification when an article is removed from the collection.

https://github.com/Financial-Times/content-collection-unfolder/pull/13


MihaiSorinMoisa:
[Removing an article from a ContentPackage causes an article delete notification to be sent](https://trello.com/c/6znPCg7d/381-removing-an-article-from-a-contentpackage-causes-an-article-delete-notification-to-be-sent)

PR #1595, opened 2018-02-06
Bridges for Coco delivery for message on K8s Publishing cluster.

yolandeleungFT:
This needs a proper eyeball review since testing it in a lower env will distort the bridge functionality as it has a consumer ID.
I left the concept bridge as it points to the vanity and therefore should continue to work in most situations.

### DIFF
--- a/cms-kafka-bridge-k8s-pub-prod-eu@.service
+++ b/cms-kafka-bridge-k8s-pub-prod-eu@.service
@@ -17,7 +17,7 @@ ExecStartPre=/bin/bash -c 'docker history coco/coco-kafka-bridge:$DOCKER_APP_VER
 ExecStart=/bin/sh -c '\
   export APP_PORT=8080; \
   export QUEUE_PROXY_ADDRS="https://upp-prod-publish-eu.ft.com/__kafka-rest-proxy"; \
-  export GROUP_ID=$(/usr/bin/etcdctl get /ft/config/environment_tag)"-kafka-bridge-pub-prod-uk"; \
+  export GROUP_ID=$(/usr/bin/etcdctl get /ft/config/environment_tag)"-kafka-bridge-k8s-pub-prod-eu"; \
   export AUTHORIZATION_KEY="Basic $(/usr/bin/etcdctl get /ft/_credentials/kafka-bridge-pub-prod/authorization_key | tr --delete \"\n\" | base64)"; \
   /usr/bin/docker run --rm --name %p-%i_$(uuidgen) -p $APP_PORT \
   --memory="256m" \
--- a/cms-kafka-bridge-k8s-pub-prod-eu@.service
+++ b/cms-kafka-bridge-k8s-pub-prod-eu@.service
@@ -1,0 +1,40 @@
+[Unit]
+Description=CMS Kafka Bridge K8s Publishing Cluster Prod UK
+After=vulcan.service
+Requires=docker.service
+Wants=cms-kafka-bridge-k8s-pub-prod-eu-sidekick@%i.service
+
+[Service]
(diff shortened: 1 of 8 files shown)


cms-kafka-bridge-k8s-pub-prod-eu@.service
```diff
@@ -0,0 +1,40 @@
...
+KillMode=none
+ExecStartPre=-/bin/bash -c '/usr/bin/docker kill "$(docker ps -q --filter=name=%p-%i_)" > /dev/null 2>&1'
+ExecStartPre=-/bin/bash -c '/usr/bin/docker rm "$(docker ps -q --filter=name=%p-%i_)" > /dev/null 2>&1'
+ExecStartPre=/bin/bash -c 'docker history coco/coco-kafka-bridge:$DOCKER_APP_VERSION > /dev/null 2>&1 \
+  || docker pull coco/coco-kafka-bridge:$DOCKER_APP_VERSION'
+
+ExecStart=/bin/sh -c '\
+  export APP_PORT=8080; \
+  export QUEUE_PROXY_ADDRS="https://upp-prod-publish-eu.ft.com/__kafka-rest-proxy"; \
+  export GROUP_ID=$(/usr/bin/etcdctl get /ft/config/environment_tag)"-kafka-bridge-pub-prod-uk"; \
```
sbuliarca: I think the GROUP_ID should be unique per bridge, and there is already one that uses this: cms-kafka-bridge-pub-prod-uk . We should put `k8s-eu` somewhere in it.

cms-kafka-bridge-k8s-pub-prod-us@.service
```diff
@@ -0,0 +1,40 @@
...
+KillMode=none
+ExecStartPre=-/bin/bash -c '/usr/bin/docker kill "$(docker ps -q --filter=name=%p-%i_)" > /dev/null 2>&1'
+ExecStartPre=-/bin/bash -c '/usr/bin/docker rm "$(docker ps -q --filter=name=%p-%i_)" > /dev/null 2>&1'
+ExecStartPre=/bin/bash -c 'docker history coco/coco-kafka-bridge:$DOCKER_APP_VERSION > /dev/null 2>&1 \
+  || docker pull coco/coco-kafka-bridge:$DOCKER_APP_VERSION'
+
+ExecStart=/bin/sh -c '\
+  export APP_PORT=8080; \
+  export QUEUE_PROXY_ADDRS="https://upp-prod-publish-us.ft.com/__kafka-rest-proxy"; \
+  export GROUP_ID=$(/usr/bin/etcdctl get /ft/config/environment_tag)"-kafka-bridge-pub-prod-uk"; \
```
sbuliarca: This should be `-kafka-bridge-pub-prod-us`, but let's also make it unique, as that is already taken

cms-kafka-bridge-k8s-pub-prod-eu@.service
yolandeleungFT: good spot.

khatton-ft: Doesn't that come from the environment_tag though? So in fact it will be `prod-uk[/us]-kafka-bridge-pub-prod-uk`?

cms-kafka-bridge-k8s-pub-prod-us@.service
```diff
@@ -0,0 +1,40 @@
+[Unit]
+Description=CMS Kafka Bridge K8s Publishing Cluster Prod UK
```
sbuliarca: Us should be in the description.

sbuliarca:
The existing bridges look good to me now.

Should we also bridge concepts in coco?
Will the services.yaml be updated at a later time?

